PR #3, opened 2022-05-17
Add first draft for Irene

gonzaponte:
This PR adds the documentation for the city *Irene* and establishes the template (set by @halmamol) for forthcoming entries.
The built documentation can be found [here](https://pr003-sw-docs.readthedocs.io/en/latest/irene.html).
On top of the documentation, there are some other changes to the rest of the docs:
- The IC file describes the arguments that are common to all cities
- The display width is now 75% of the full page width
- The toc in the IC file now has a max depth of 1, hiding the subsections within each city entry.


docs/source/irene.rst
```diff
@@ -1,5 +1,193 @@
 Irene
-==========
+=====
 
-.. caution::
-  This page is currently not created, we are working on it. If you would like to contribute on this documentation, reach `me <helena.almamol@gmail.com>`_!
+*From ancient Greek, Εἰρήνη, meaning peace.*
+
+The S1 and S2 signals are localized to relatively short time intervals within the sensors' waveforms. Thus, *Irene* processes **RWF**\ s to find these time slices (peaks) and disregard the rest of the waveform. During this procedure, PMT and SiPM waveforms are matched and combined into a single structure. The collection of all peaks in an event is called a Peak-map or **PMap**.
```
halmamol: Don't forget to specify here from where the Irene inputs (RWFs) come from.

docs/source/irene.rst
```diff
@@ -1,5 +1,193 @@
...
+ * ``/PMAPS/S2Si``: the sliced individual SiPM waveforms for each S2 peak. 4 columns: event number, peak number, sipm id and amplitude (pes)
+ * ``/Filters/empty_pmap``: flag for whether an event passed the empty pmap filter
+ * ``/Filters/s12_indices``: flag for whether an event passed the s12 indices filter
+
+.. _Irene config:
+
+Config
+------
+
+Besides the :ref:`Common arguments to every city`, *Irene* accepts the following arguments:
```
halmamol: Is _accepts_ the best word? They're required, isn't it? Or you mean that if you don't give them it would use the default parameters?

docs/source/irene.rst
```diff
@@ -1,5 +1,193 @@
...
+     - ``float``
+     - Sampling period of SiPMs. Should be removed.
+
+
+.. _Irene workflow:
+
+Workflow
+--------
+
+Irene performs a number of data transformations in order to obtain a **PMap**. These operations can be grouped in four main tasks:
```
halmamol: Are these tasks done in this order?

docs/source/irene.rst
```diff
@@ -1,5 +1,193 @@
...
+
+The resulting waveform is still bipolar. This is addressed by the deconvolution algorithm (BLR). This process is fairly complex, but in simple terms, it consists of a high-pass filter and a signal accumulator, which inverts the effect of the PMT electronics. For greater detail on the PMT electronics and the recovery algorithm see [1]_. Finally, the polarity of the waveform is inverted to make it positive.
+
+All the aforementioned steps are performed for each PMT separately. The output of this algorithm are the so-called *Corrected waveforms* (**CWF**\ s).
+
+The city *Isidora* allows the user to run just this stage of the reconstruction and store the **CWF**\ s for further study. Irene however, does not store them and they are fed directly into the rest of the PMap-building algorithm. The **CWF** corresponding to the **RWF** shown above is:
```
halmamol: Add link to Isidora

docs/source/irene.rst
```diff
@@ -1,5 +1,193 @@
...
+ .. image:: images/irene/sipm_rwf.png
+   :width: 850
+
+
+.. _Waveform calibration:
+
+Waveform calibration
+::::::::::::::::::::
+
+The production and manufacturing of the sensors and other electronic components does not guarantee a homogeneous response among all sensors. Thus, the waveforms are calibrated to equalize their response. The calibration consists of a constant for each sensor indicating the number of ADC corresponding to a photoelectron (calibration constant), which is a physical quantity common to all of them. The calibration technique is similar for PMTs and SiPMs. For details about the calibration procedure see <reference to calibration procedures>.
```
halmamol: Reference here missing

docs/source/irene.rst
```diff
@@ -1,5 +1,193 @@
...
+
+The calibration constants are measured regularly while the detector is in operation. The calibration constants are fetched from the database automatically and indexed by run number.
```
halmamol, line 152: Maybe add reference here to calibration cities? We can also add it in the future

docs/source/irene.rst
gonzaponte: I am not sure whether _accepts_ is appropriate or not, but it's probably not the best term, as you point out.

docs/source/irene.rst
gonzaponte: The first two don't need to follow that specific order, but the implementation follows that sequence. Apart from that, the order needs to be that one because each step relies on the transformations of the previous ones.

docs/source/irene.rst
gonzaponte: We don't have a link to put here, I'm afraid... What should we do?

docs/source/irene.rst
halmamol: I would then specify that the city follows that sequence, for compliteness.

docs/source/irene.rst
halmamol: You refer here about PMTs/SiPMS calibrations? Maybe we can add some reference to any (or all) the calibration cities (Phyllis/Trude/Beredice).

halmamol: It could be in connection with the following comment (the one about the calibration cities). We can update that part in the future.

halmamol (approved):
This PR adds the first city (Irene) in the documentation, and corrects some things along the readthedocs. This city can be used as template for the following ones. Good job @gonzaponte! Approved!

PR #5, opened 2022-07-13
Add first draft of Isaura documentation

ausonandres:
This PR adds the documentation for the city Isaura, following the same structure as for Irene, written by @gonzaponte.

You can check the version of the city doc here: https://pr005-sw-docs.readthedocs.io/en/latest/isaura.html


docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
 Isaura
-==========
+======
 
-.. caution::
-  This page is currently not created, we are working on it. If you would like to contribute on this documentation, reach `me <helena.almamol@gmail.com>`_!
+*From ancient greek, Ισαυρία: an ancient rugged region in Asia Minor.*
+
+This city just computes tracks and extracts topology information from **deconvoluted hits**. Therefore, it is analogous to the final stage of :doc:`esmeralda`. Unlike the latter, the input hits correspond to the deconvoluted ones coming from :doc:`beersheba` (instead of :doc:`penthesilea`), allowing us to obtain a finer tracking of events.
```
paolafer: I would drop `just` from this sentence, because it sounds informal.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+.. _Isaura output:
+
+Output
+------
+
+ * ``/Tracking/Tracks``: tracking-related information of events (in particular: track energy (``energy``), track length (``length``), number of voxels (``numb_of_voxels``), number of hits (``numb_of_hits``), minimum and maximum position computed with the hits comprising the track (``x_min``, ``y_max``, ...), blobs position (``blob1_z``, ``blob2_x``,...) and energy (``eblob1``, ``eblob2``), energy of the hits shared by both blobs (``ovlp_blob_energy``), and voxel size (``vox_size_x``, ``vox_size_y``, ``vox_size_z``). Each row corresponds to a different track, specified among the others within an event with its ``trackID``
+ * ``/Summary/Events``: global information related to the event. Each row is one event
+ * ``/DST/Events``: copy of the point-like information (*kdst*) of events, output of :doc:`penthesilea`
+ * ``/Filters/hits_select``: flag to indicate if an event did not pass due to the lack of hits
+ * ``/Filters/topology_select``: flag that indicates whether the topology has been performed properly
```
paolafer: What does it mean that the topology performs properly? I think we need to be a bit clearer.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+Output
+------
+
+ * ``/Tracking/Tracks``: tracking-related information of events (in particular: track energy (``energy``), track length (``length``), number of voxels (``numb_of_voxels``), number of hits (``numb_of_hits``), minimum and maximum position computed with the hits comprising the track (``x_min``, ``y_max``, ...), blobs position (``blob1_z``, ``blob2_x``,...) and energy (``eblob1``, ``eblob2``), energy of the hits shared by both blobs (``ovlp_blob_energy``), and voxel size (``vox_size_x``, ``vox_size_y``, ``vox_size_z``). Each row corresponds to a different track, specified among the others within an event with its ``trackID``
+ * ``/Summary/Events``: global information related to the event. Each row is one event
+ * ``/DST/Events``: copy of the point-like information (*kdst*) of events, output of :doc:`penthesilea`
+ * ``/Filters/hits_select``: flag to indicate if an event did not pass due to the lack of hits
+ * ``/Filters/topology_select``: flag that indicates whether the topology has been performed properly 
+ * ``MC info``: Monte Carlo information of events. Only if run number < 0
```
paolafer: What is the path of these tables?

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+     - ``int``
+     - The voxel dropping procedure commented on ``energy_threshold`` can only happen if the number of voxels is larger than the value specified in this argument.
+
+   * - ``blob_radius``
+     - ``float``
+     - Radius of the blobs.
+
+   * - ``max_num_hits``
+     - ``int``
+     - Maximum number of hits for an event to be processed.
```
paolafer, line 72: What happens if an event has more hits than this value? Is it filtered out? If that's the case, you should add this information to the ``/Filters/hits_select`` documentation above (meaning, an event can be filtered out either if it has no hits, and if it has too many hits).

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+The XY (d), XZ (e) and YZ (f) projections of devonvoluted hits, along with the blobs computed with this algortihm, for the same event as the one shown before (in the voxelization plot) can be seen above. This image illustrates how the blobs seem to be computed perfectly. According to our reconstruction, it corresponds to a clear single-electron event (background), due to the noticeable difference between the energy of its blobs: ``eblob1`` = 755 keV, whereas ``eblob2`` = 104 keV.
+
+           
+`Isaura` comprises the last step within the NEXT reconstruction chain. Therefore, after it, we have access to all the relevant information to perform the analysis. This information is finally stored in different tables, just as the :ref:`Output <Isaura output>` subsection indicates.
+
+
+
+
+ .. [#] When regarding the plot, it is important to realize that the hits considered are the ones from :doc:`penthesilea`. Events coming from :doc:`beersheba` comprise a much larger amount of *deconvoluted hits* (more than one order of magnitude), given the finer granularity.
```
paolafer: Probably `when` is superfluous here: "Regarding the plot...".

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+.. _Isaura workflow:
+
+Workflow
+--------
+
+The basic idea behind Isaura is to run the "traditional" *Paolina* algorithm over the deconvoluted hits output in :doc:`beersheba`, in order to convert ``events`` into ``tracks`` with ``blobs``.
+
+The first step within the algorithm consists in checking that the number of hits is lower than the value provided in the config file (``max_num_hits``). That argument was introduced because, when running *Paolina* algorithm after :doc:`penthesilea`, there were some events that comprise such large amount of hits that the tracking information extraction took ridiculously long. As the plot presented below shows [#]_, high energy (trigger2) events usually contain around 200 *Penthesilea hits*, while there are some of them with more than 10000 hits.
+
+ .. image:: images/isaura/nhits_per_evt_r8571.jpg
```
paolafer, line 104: What is the difference between the left and right plot? Please, explain.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+   :width: 1000
+
+The plot also shows that these events only  appear a few times within a 24h-long low-background run (around 0.05% of the total set of events). Rejecting these type of events is not a particularly worrysome issue: they would be thrown away in the posterior analysis, seeing that none of them are exclusively contained inside the fiducial volume. The spatial distributions for one of these events is presented below (where the grey dashed lines illustrate the boundaries of the chamber):
```
paolafer: `seeing that none of them are exclusively contained` --> because none of them are exclusively contained. It sounds more formal to me.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+
+ .. image:: images/isaura/XY_Z_distributions_evt_many_hits.jpg
+   :width: 1000
+
+In any case, one can easily infer from the plots that these events are not physical. On the contrary, they seem to correspond to either some kind of flash occurring in the chamber (like a mini-spark) or some fail in the electronics (after the saturation of an alpha particle, for example). The ID of the events that are removed from the reconstruction chain because of this reason will be specified in the table ``Filters/topology_select``, in order to keep track of this information.
+
+Another obvious condition that must fulfill all events to be processed is to contain hits. If not, the event will be also rejected, which will be displayed in the table ``Filters/hits_select``.
```
paolafer: `that must fulfill all events` --> that all events must fulfill.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+..
+ Next step includes another (quite obvious) check: at least one hit inside the event must have a well-defined energy. If not, the event will be also rejected, since no topological information could be extracted.
+
+
+.. _Exracting the topology-related information:
+
+Extracting the topology-related information
+:::::::::::::::::::::::::::::::::::::::::::
+As it is well known, an excellent topological discrimation between signal and background (thanks to the usage of a gaseous medium inside the TPC) is one of the fundamental trademarks of the NEXT experiment. In order to do that, it is necessary to (1) separate the different tracks that may form the event, (2) find the extremes for each of them, and (3) compute the energy around these extremes, providing the so-called *blobs*.
```
paolafer: `discrimation` --> discrimination.

paolafer: Remove `As it is well known`, it's superfluous and a bit arrogant in my opinion ;-).

paolafer: `In order to do that` --> In order to achieve that. You are referring to the topological discrimination, which you don't _do_, but rather _achieve_.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+In order to compute all the tracking information presented below it will be mandatory first to check that every event contain hits with well-defined energy. For instance, events with all hits outside the krypton correction map bondaries will be thrown away, since their energy cannot be corrected and their ``Ec`` variable (*corrected energy*) will be ``NaN``.
```
paolafer: `below` --> above (I guess..). Or you can just say "To compute this tracking features".

docs/source/isaura.rst
ausonandres: Sorry, I don't understand the question... Are you asking about the MC tables?

docs/source/isaura.rst
ausonandres: Yes, it is filtered out, but it is indicated in `/Filters/topology_select`... i can add that information, anyway.

docs/source/isaura.rst
paolafer: I was wondering why for the other tables you specify the path `/Filters/topology_select` and for the MC info you don't (`MC info`).

docs/source/isaura.rst
paolafer: I see, I thought that they were counted in `/Filters/hits_select`. That's ok, then.

docs/source/isaura.rst
paolafer: `bondaries` --> boundaries.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+ .. image:: images/isaura/r8250_evt194237_dhits.png
+   :width: 46%
+ .. image:: images/isaura/r8250_evt194237_voxels.png
+   :width: 53%
+
+
+Seeing that events have already been divided into voxel-made 3D tracks, the next (and final) step comprises the computation of the **blobs** [#]_. This is a crucial stage within the reconstruction chain, since the energy of these elements will allow us to distinguish between double-electron (such as the double-beta signal) and single-electron (the majority of backgrounds) tracks.
+
+There are different ways to define the center of the blobs, the one selected by NEXT is illustrated in the figure below. According to this procedure, the two extreme voxels of the track are localized following the BFS algorithm, on a first stage. The energy-weighted averaged position of the hits inside the voxel will correspond to the **blob center**. From that point, a 3D sphere of radius ``blob_radius`` (specified in the config file) is taken, and all hits (independent to the voxel they belong to) inside that sphere will contribute to the energy of the blob.
```
paolafer: `independent to` --> independent of.

paolafer: `There are different ways to define the center of the blobs, the one selected by NEXT is illustrated in the figure below...` --> I think it's more direct to say something on the line of "To define the center of the blobs, we use the following procedure, as illustrated in the figure below. First,  the two extreme voxels of the track are localized following the BFS algorithm."  `on a first stage` sounds strange to me (I think it would be "in a first stage" anyway).

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+.. _Isaura display:
+
+ .. image:: images/isaura/r8250_evt194237_dhits.png
+   :width: 46%
+ .. image:: images/isaura/r8250_evt194237_voxels.png
+   :width: 53%
+
+
+Seeing that events have already been divided into voxel-made 3D tracks, the next (and final) step comprises the computation of the **blobs** [#]_. This is a crucial stage within the reconstruction chain, since the energy of these elements will allow us to distinguish between double-electron (such as the double-beta signal) and single-electron (the majority of backgrounds) tracks.
+
```
paolafer, line 161: I would describe how to find the extremes of the track first, and, later, how to calculate the blob energy. It seems the logical order to me. In the current version you have to say `To carry out this procedure, and as stated before,` which seems a bit winding.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+ - It is possible that some spurious **low-energy** hits appear around the track (due to over-iterations during the *Richarson-Lucy* deconvolution process, as commented in :doc:`beersheba`; or some noise inside the chamber, for example). If these hits are reconstructed around the track but not far enough to produce a different S2 or track (taking into account the voxel size), they can be considered as a part of the main one and, being a bit separate, it is probable they end up belonging to an extreme voxel. That case would not be correct, and in order to solve it, the voxel will be dropped from the track and its energy passed to the closest one. This process is only carried out if the voxel energy is lower than: ``energy_threshold`` and the track is made by more than ``min_voxels`` voxels. Once this procedure is done, the extreme voxels are searched and found again recursively, until none of these conditions are fulfilled.
```
paolafer: *Richarson-Lucy* --> Richardson-Lucy.

paolafer: `it is probable they end up` --> it is probable that they end up. I think that in this case you need the `that`.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+ - Another particular scenario is the one that comes up when there are multiple end-voxel candidates (one can imagine that the shorter the track the more probable is this to happen). To deal with it, the more energetic candidates will be the ones set as extremes. With this convention, we aim to minimize the voxel-dropping algorithm commented just above.
```
paolafer: It's not clear to me how this convention helps minimizing the voxel dropping.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+
+Eventually, once the extreme voxels are properly found, the computation of the center and energy of the blobs (stored in the ``Tracks/Tracking`` table as: ``blobi_x``, ``blobi_y``, ``blobi_z``, and ``eblobi`` (with ``i`` being 1, 2), respectively) will be rather straightforward, just as the previous figure pointed out. The only comment to take into account here is that not every hit falling inside the blob sphere will be considered for its total energy, but only the ones that belong to an adjacent voxel.
```
paolafer: `only the ones that belong to an adjacent voxel`: adjacent to what? To the end-point voxel? Say it explicitly.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+As convention, it is worth-remarking that the blob1 and blob2 are defined in such a way that they fulfill: ``eblob1`` > ``eblob2``.
```
paolafer: `worth-remarking`: do you need the dash? Isn't it better "it is worth remarking that..."?

paolafer: I would say more explicitly: "... that the assignment of `1` and `2` is done in such a way that ``eblob1`` > ``eblob2``."

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+`Isaura` comprises the last step within the NEXT reconstruction chain. Therefore, after it, we have access to all the relevant information to perform the analysis. This information is finally stored in different tables, just as the :ref:`Output <Isaura output>` subsection indicates.
+
+
+
+
+ .. [#] When regarding the plot, it is important to realize that the hits considered are the ones from :doc:`penthesilea`. Events coming from :doc:`beersheba` comprise a much larger amount of *deconvoluted hits* (more than one order of magnitude), given the finer granularity.
+
+ .. [#] T. H. Cormen, C. Stein, R. L. Rivest, and C. E. Leiserson, Introduction to algorithms. McGraw-Hill Higher Education, 2nd ed., 2001.
+
+ .. [#] For those who don't know, these famous *blobs* are imaginary 3D spheres located around both ends of each track. Their energy (provided by the hits contained within the sphere) corresponds to an excellent tool to investigate whether there has been a large and sudden energy deposition in the track extreme (i.e. *Bragg peak*, indicating the stopping point of a charged particle) or not (starting point of its trajectory).
```
paolafer: This sentence is too informal in my opinion. I would simply say something like "The blobs are imaginary 3D spheres located around both ends of each track. Their energy (provided by the hits contained within the sphere) is an excellent tool to investigate whether there has been a large and sudden energy deposition in the track extreme (i.e. *Bragg peak*, indicating the stopping point of a charged particle) or not (starting point of its trajectory)."

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+
+Eventually, once the extreme voxels are properly found, the computation of the center and energy of the blobs (stored in the ``Tracks/Tracking`` table as: ``blobi_x``, ``blobi_y``, ``blobi_z``, and ``eblobi`` (with ``i`` being 1, 2), respectively) will be rather straightforward, just as the previous figure pointed out. The only comment to take into account here is that not every hit falling inside the blob sphere will be considered for its total energy, but only the ones that belong to an adjacent voxel.
+
+As convention, it is worth-remarking that the blob1 and blob2 are defined in such a way that they fulfill: ``eblob1`` > ``eblob2``.
+
+
+The final step of the *Paolina* algorithm includes the computation of the ``ovlp_blob_energy`` variable. In short tracks it is common to have **overlapping blobs**, i.e. blobs that share some of their hits [#]_. Whenever this happen, the blob energies will not be correctly computed, due to an over-estimation of the total energy of the track. This variable will materialize this effect, providing a really usefull tool to get rid of these cases.
```
paolafer: `usefull` --> useful.
`This variable will materialize this effect`: I would use the present tense, but changing the verb (materialize sounds odd to me). However, you haven't defined this variable yet, right? I would first define the variable.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+
+ .. image:: images/isaura/RunVI_b_evt_1720keV_XYZ.jpg
+   :width: 1200
+
+The XY (d), XZ (e) and YZ (f) projections of devonvoluted hits, along with the blobs computed with this algortihm, for the same event as the one shown before (in the voxelization plot) can be seen above. This image illustrates how the blobs seem to be computed perfectly. According to our reconstruction, it corresponds to a clear single-electron event (background), due to the noticeable difference between the energy of its blobs: ``eblob1`` = 755 keV, whereas ``eblob2`` = 104 keV.
```
paolafer: `The XY (d), XZ (e) and YZ (f) projections of devonvoluted hits, along with the blobs computed with this algortihm, for the same event as the one shown before (in the voxelization plot) can be seen above.` --> "The XY (d), XZ (e) and YZ (f) projections of **deconvoluted** hits, along with the blobs computed with this **algorithm** can be seen in the above picture, for the same event as the one shown in the voxelization plot". You can consider adding numbers to the plots to refer to them.

docs/source/isaura.rst
```diff
@@ -1,5 +1,160 @@
...
+
+           
+`Isaura` comprises the last step within the NEXT reconstruction chain. Therefore, after it, we have access to all the relevant information to perform the analysis. This information is finally stored in different tables, just as the :ref:`Output <Isaura output>` subsection indicates.
```
paolafer: Remove `just`, too informal.

docs/source/isaura.rst
ausonandres: Well, that paragraph just intended to present the way we define the blobs. But both the procedure to find the extremes and the computation of the blob energy is explained below, following the order that you claimed.

docs/source/isaura.rst
ausonandres: if you have two candidates for the extremes: one below the `energy_threshold` variable and the other one above that, the voxel dropping algorithm will not be performed, because the latter will be the one selected as the extreme voxel, according to the convention.

docs/source/isaura.rst
paolafer: My comment comes from the fact that the explanation of how the blob energy is calculated is very detailed and it seems that belongs to the explanation that comes later. Here you can just leave the sentence `the next (and final) step comprises the computation of the blobs [3]. This is a crucial stage within the reconstruction chain, since the energy of these elements will allow us to distinguish between double-electron (such as the double-beta signal) and single-electron (the majority of backgrounds) tracks.`, which introduces the blobs and ends the general overview of the topological algorithms, before you explain them in detail. At least, this is how I understand the flow of the documentation.

docs/source/isaura.rst
paolafer: I would state this explicitly.

docs/source/isaura.rst
```diff
@@ -1,5 +1,163 @@
...
+
+ - Another particular scenario is the one that comes up when there are multiple end-voxel candidates (one can imagine that the shorter the track the more probable is this to happen). To deal with it, the more energetic candidates will be the ones set as extremes. With this convention, we aim to minimize the voxel-dropping algorithm commented just above.
+
+
+Eventually, once the extreme voxels are properly found, the computation of the center and energy of the blobs (stored in the ``Tracks/Tracking`` table as: ``blobi_x``, ``blobi_y``, ``blobi_z``, and ``eblobi`` (with ``i`` being 1, 2), respectively) will be rather straightforward, just as the previous figure pointed out. The only comment to take into account here is that not every hit falling inside the blob sphere will be considered for its total energy, but only the ones that belong to a voxel adjacent to the extreme.
+
+As convention, the assignment of ``1`` and ``2`` in defined in such a way that ``eblob1`` > ``eblob2``.
```
paolafer: `in defined` --> is defined.

docs/source/isaura.rst
ausonandres: Ok, I have just changed now the structure of this part, trying to be more similar to what you proposed. Let me know it that works for you.

docs/source/isaura.rst
```diff
@@ -1,5 +1,163 @@
...
+ * ``/Summary/Events``: global information related to the event. Each row is one event.
+ * ``/DST/Events``: copy of the point-like information (*kdst*) of events, output of :doc:`penthesilea`.
+ * ``/Filters/hits_select``: flag to indicate if an event did not pass due to the lack of hits.
+ * ``/Filters/topology_select``: flag that indicates whether events have too many hits according to the value specified in the config file. If so, its ID will not appear in the table and the topology information will not be computed.
+ * MC info: copy of the Monte Carlo information for the events that output the city. Only if run number < 0. The tables included are: ``/MC/configuration``, ``/MC/hits``, ``/MC/particles``, ``/MC/sns_positions``, and ``/MC/sns_response``.
```
paolafer: `copy of the Monte Carlo information for the events that output the city`: what do you mean with "events that output the city"? Events that appear in the output of the city? In that case, I think it should be "for the events that pass the filters of the city" or "for the events that the city outputs", if you use `to output` as a synonymous of `to generate`.

docs/source/isaura.rst
```diff
@@ -1,5 +1,163 @@
...
+   :widths: 50 100 120
+   :header-rows: 1
+
+   * - **Parameter**
+     - **Type**
+     - **Description**
+
+   * - ``vox_size``
+     - ``[float,float,float]``
+     - X, Y, and Z dimensions of the voxels used in the voxalization of the hits.
```
paolafer: `voxalization` --> voxelization.

docs/source/isaura.rst
```diff
@@ -1,5 +1,163 @@
...
+
+ .. image:: images/isaura/nhits_per_evt_r8571.jpg
+   :width: 1000
+
+The plot also shows that these events only  appear a few times within a 24h-long low-background run (around 0.05% of the total set of events). Rejecting these type of events is not a particularly worrysome issue: they would be thrown away in the posterior analysis, owing to the fact that none of them are exclusively contained inside the fiducial volume. The spatial distributions for one of these events is presented below (where the grey dashed lines illustrate the boundaries of the chamber):
```
paolafer: `these type` --> either "this type" or "these types".

docs/source/isaura.rst
```diff
@@ -1,5 +1,163 @@
...
+
+
+ - It is possible that some spurious **low-energy** hits appear around the track (due to over-iterations during the Richarson-Lucy deconvolution process, as commented in :doc:`beersheba`; or some noise inside the chamber, for example). If these hits are reconstructed around the track but not far enough to produce a different S2 or track (taking into account the voxel size), they can be considered as a part of the main one and, being a bit separate, it is probable that they end up belonging to an extreme voxel. That case would not be correct, and in order to solve it, the voxel will be dropped from the track and its energy passed to the closest one. This process is only carried out if the voxel energy is lower than: ``energy_threshold`` and the track is made by more than ``min_voxels`` voxels. Once this procedure is done, the extreme voxels are searched and found again recursively, until none of these conditions are fulfilled.
```
paolafer: `is lower than:`. I think that `:` is a typo, right?

docs/source/isaura.rst
```diff
@@ -1,5 +1,163 @@
...
+
+ - Another particular scenario is the one that comes up when there are multiple end-voxel candidates (one can imagine that the shorter the track the more probable is this to happen). To deal with it, the more energetic candidates will be the ones set as extremes. With this convention, we aim to minimize the voxel-dropping algorithm commented just above.
+
+
+Eventually, once the extreme voxels are properly found, the computation of the center and energy of the blobs (stored in the ``Tracks/Tracking`` table as: ``blobi_x``, ``blobi_y``, ``blobi_z``, and ``eblobi`` (with ``i`` being 1, 2), respectively) will be rather straightforward, just as the previous figure pointed out. The only comment to take into account here is that not every hit falling inside the blob sphere will be considered for its total energy, but only the ones that belong to a voxel adjacent to the extreme.
```
paolafer: `(with ``i`` being 1, 2), respectively)`: remove one closing parenthesis.

docs/source/isaura.rst
```diff
@@ -1,5 +1,163 @@
...
+
+As convention, the assignment of ``1`` and ``2`` in defined in such a way that ``eblob1`` > ``eblob2``.
+
+
+The final step of the *Paolina* algorithm includes the computation of the ``ovlp_blob_energy`` (“*overlap blob energy*”) variable: in short tracks it is common to have **overlapping blobs**, i.e. blobs that share some of their hits [#]_. In these cases, the blob energies are not computed correctly, since they would produce an over-estimation of the total energy of the track. Therefore, it will be interesting to reject this type of events during the posterior analysis in case the blob energy distributions want to be studied. Owing to the fact that ``ovlp_blob_energy`` is defined as the total amount of energy of these shared hits, a selection of ``ovlp_blob_energy = 0`` will get rid of these events easily.
```
paolafer: ` In these cases, the blob energies are not computed correctly, since they would produce an over-estimation of the total energy of the track.`: I'm not sure I understand the sentence correctly. One can calculate the blob energies not correctly if a hit belongs to both blobs, but the total energy of the track is still computed correctly, because it is not calculated from the blob energy, right?

docs/source/isaura.rst
ausonandres: ok

docs/source/isaura.rst
paolafer: Ok!

docs/source/isaura.rst
ausonandres: the parenthesis was opened before: `(stored in the ``Tracks/Tracking`` table as: ``blobi_x``, ``blobi_y``, and ``blobi_z``, (with ``i`` being 1, 2), respectively)`. Anyway, I can change it to `-` if you prefer.

paolafer: I see. Yes, I think that using different symbols for the two inserts would be clearer.

docs/source/isaura.rst
ausonandres: Yes, you are right. It is obvious that the energy of the blob does not affect the total energy of events, but what I meant is that if there is some overlapping, the total energy of the blobs makes no sense because it can be even larger than the total energy of the event. I'll rephrase it.

paolafer (approved):
This PR adds the documentation of Isaura. It is very detailed, clear and with useful examples. Excellent job, @ausonandres!

docs/source/isaura.rst
```diff
@@ -1,5 +1,168 @@
...
+ * ``/DST/Events``
+
+.. _Isaura output:
+
+Output
+------
+
+ * ``/Tracking/Tracks``: tracking-related information of events (in particular: track energy (``energy``), track length (``length``), number of voxels (``numb_of_voxels``), number of hits (``numb_of_hits``), minimum and maximum position computed with the hits comprising the track (``x_min``, ``y_max``, ...), blobs position (``blob1_z``, ``blob2_x``,...) and energy (``eblob1``, ``eblob2``), energy of the hits shared by both blobs (``ovlp_blob_energy``), and voxel size (``vox_size_x``, ``vox_size_y``, ``vox_size_z``). Each row corresponds to a different track, specified among the others within an event with its ``trackID``.
+ * ``/Summary/Events``: global information related to the event. Each row is one event.
+ * ``/DST/Events``: copy of the point-like information (*kdst*) of events, output of :doc:`penthesilea`.
```
halmamol: Could you write kdst in bold letters instead of italic?

docs/source/isaura.rst
```diff
@@ -1,5 +1,168 @@
...
+ * ``/Filters/hits_select``: flag to indicate if an event pass the selection of having more than 0 hits. 
+ * ``/Filters/topology_select``: flag to indicate if an event pass the selection of having less hits than the ``max_num_hits`` parameter specified in the config file.
+ * MC info: copy of the Monte Carlo information for the events that the city outputs. Only if run number < 0. The tables included are: ``/MC/configuration``, ``/MC/hits``, ``/MC/particles``, ``/MC/sns_positions``, and ``/MC/sns_response``.
```
halmamol: run number -> ``run_number``

docs/source/isaura.rst
```diff
@@ -1,5 +1,168 @@
...
+
+
+.. _Isaura workflow:
+
+Workflow
+--------
+
+The basic idea behind Isaura is to run the "traditional" *Paolina* algorithm over the deconvoluted hits output in :doc:`beersheba`, in order to convert ``events`` into ``tracks`` with ``blobs``.
+
+The first step within the algorithm consists in checking that the number of hits is lower than the value provided in the config file (``max_num_hits``). That argument was introduced because, when running *Paolina* algorithm after :doc:`penthesilea`, there were some events that comprise such large amount of hits that the tracking information extraction took ridiculously long. The following picture shows the number of Penthesilea hits per event (with a different scale) for a typical 24h-long low-background run included in the NEXT-White double-beta analysis [#]_. High energy (trigger2) events usually contain around 200 *Penthesilea hits* (as right panel points out), while there are some of them with more than 10000 hits (illustrated in left panel).
```
halmamol: Several comments on this paragraph:
- I would add a title to this initial step in the city (it would be one of the tasks listed at the beginning of the workflow - see a comment that I have added later about the tasks) 
- Add "**(hdst)**" after "*Penthesilea hits*"
- Any reason of using 'panel' instead of 'plots'?

docs/source/isaura.rst
```diff
@@ -1,5 +1,168 @@
 Isaura
-==========
+======
 
-.. caution::
-  This page is currently not created, we are working on it. If you would like to contribute on this documentation, reach `me <helena.almamol@gmail.com>`_!
+*From ancient greek, Ισαυρία: an ancient rugged region in Asia Minor.*
+
+This city computes tracks and extracts topology information from **deconvoluted hits**. Therefore, it is analogous to the final stage of :doc:`esmeralda`. Unlike the latter, the input hits correspond to the deconvoluted ones coming from :doc:`beersheba` (instead of :doc:`penthesilea`), allowing us to obtain a finer tracking of events.
```
halmamol: Can you add "**(ddst)**" right after 'deconvoluted hits'?

docs/source/isaura.rst
```diff
@@ -1,5 +1,168 @@
...
+
+       Events with more hits are filtered out and their ID will not appear in the ``Filters/topology_select`` table of the output.
+
+
+.. _Isaura workflow:
+
+Workflow
+--------
+
+The basic idea behind Isaura is to run the "traditional" *Paolina* algorithm over the deconvoluted hits output in :doc:`beersheba`, in order to convert ``events`` into ``tracks`` with ``blobs``.
```
halmamol: I would suggest to list at the beginning of the section the tasks that the city does (in the same manner that it's done in Irene - see screenshot below). In this way it will be easier to explore the different task on the webpage and it's not necessary to be scrolling around: 
imagen

docs/source/isaura.rst
```diff
@@ -1,5 +1,168 @@
...
+
+..
+ Next step includes another (quite obvious) check: at least one hit inside the event must have a well-defined energy. If not, the event will be also rejected, since no topological information could be extracted.
+
+
+.. _Exracting the topology-related information:
+
+Extracting the topology-related information
+:::::::::::::::::::::::::::::::::::::::::::
+An excellent topological discrimination between signal and background (thanks to the usage of a gaseous medium inside the TPC) is one of the fundamental trademarks of the NEXT experiment. In order to achieve that, it is necessary to (1) separate the different tracks that may form the event, (2) find the extremes for each of them, and (3) compute the energy around these extremes, providing the so-called *blobs*.
```
halmamol: Maybe to add these points as a list? You can even separate the following text in sub(-sub) sections

docs/source/isaura.rst
```diff
@@ -1,5 +1,168 @@
...
+ .. image:: images/isaura/r8250_evt194237_dhits.png
+   :width: 44.8%
+ .. image:: images/isaura/r8250_evt194237_voxels.png
+   :width: 53%
+
+
+Since events have already been divided into voxel-made 3D tracks, the next (and final) step involves the computation of the **blobs**. They are defined as imaginary 3D spheres located around both ends of each track. Their energy is an excellent tool to investigate whether there has been a large and sudden energy deposition in the track extreme (i.e. *Bragg peak*, indicating the stopping point of a charged particle) or not (starting point of its trajectory). Therefore, this is a crucial stage within the reconstruction chain, since that will allow us to separate double-electron (such as the double-beta signal) from single-electron (the majority of backgrounds) tracks.
+
+
+To compute the position of the blobs, we need to find the two extreme voxels of the track, which is done following the BFS algorithm. Then, the energy-weighted averaged position of the hits inside these voxels will correspond to the **blob center**, as illustrated in the figure below.
```
halmamol: Maybe to add on the figure what is 'blob center' and 'hits'?

docs/source/isaura.rst
```diff
@@ -1,5 +1,168 @@
...
+
+ - Another particular scenario is the one that comes up when there are multiple end-voxel candidates (one can imagine that the shorter the track the more probable is this to happen). To deal with it, the more energetic candidates will be the ones set as extremes. With this convention, we aim to minimize the use of the voxel-dropping algorithm commented above for those cases where the energy of one candidate is larger than ``energy_threshold`` while the other one is below that value. 
+
+
+Once the extreme voxels are properly found, the center position of the blobs --stored in the ``Tracks/Tracking`` table as: ``blobi_x``, ``blobi_y``, and ``blobi_z``, (with ``i`` being 1, 2), respectively-- is computed in accordance with the figure previously presented.
```
halmamol, line 179: I would add here the reference to the footnote about the definition of Blob 1 and Blob 2 (you do it in the following paragraph)

docs/source/isaura.rst
```diff
@@ -1,5 +1,168 @@
...
+
+ .. image:: images/isaura/nhits_per_evt_r8571.jpg
+   :width: 1000
+
+The plot also shows that these events only  appear a few times within a 24h-long low-background run (around 0.05% of the total set of events). Rejecting this type of events is not a particularly worrysome issue: they would be thrown away in the posterior analysis, owing to the fact that none of them are exclusively contained inside the fiducial volume. The spatial distributions for one of these events is presented below (where the grey dashed lines illustrate the boundaries of the chamber):
```
halmamol: I would remove the ":" at the end of the paragraph (you don't add them in any of the previous pictures)

docs/source/isaura.rst
ausonandres: As far as I was concerned, when there is a plot with different subplots they are generally referenced as 'panels', however i don't have a strong opinion, so i can change it if you want.

docs/source/isaura.rst
ausonandres: Ok, I didn't do that because there is only one thing to do in the city (compute topology information), so I believed it made no sense.

docs/source/isaura.rst
ausonandres: I would say that the figure is clear, but I'm going to add that information in the text anyway.

docs/source/isaura.rst
ausonandres: When the position of the blobs is computed, labels 1 and 2 are selected randomly. They are reassigned once energies are calculated. Therefore, I would not add the reference here, since ``eblob1`` and ``eblob2`` parameters are not defined yet.

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+.. _Isaura output:
+
+Output
+------
+
+ * ``/Tracking/Tracks``: tracking-related information of events (in particular: track energy (``energy``), track length (``length``), number of voxels (``numb_of_voxels``), number of hits (``numb_of_hits``), minimum and maximum position computed with the hits comprising the track (``x_min``, ``y_max``, ...), blobs position (``blob1_z``, ``blob2_x``,...) and energy (``eblob1``, ``eblob2``), energy of the hits shared by both blobs (``ovlp_blob_energy``), and voxel size (``vox_size_x``, ``vox_size_y``, ``vox_size_z``). Each row corresponds to a different track, specified among the others within an event with its ``trackID``.
+ * ``/Summary/Events``: global information related to the event. Each row is one event.
+ * ``/DST/Events``: copy of the point-like information (**kdst**) of events, output of :doc:`penthesilea`.
+ * ``/Filters/hits_select``: flag to indicate if an event pass the selection of having more than 0 hits.
```
tcontrer: event pass -> event passes

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+.. _Isaura output:
+
+Output
+------
+
+ * ``/Tracking/Tracks``: tracking-related information of events (in particular: track energy (``energy``), track length (``length``), number of voxels (``numb_of_voxels``), number of hits (``numb_of_hits``), minimum and maximum position computed with the hits comprising the track (``x_min``, ``y_max``, ...), blobs position (``blob1_z``, ``blob2_x``,...) and energy (``eblob1``, ``eblob2``), energy of the hits shared by both blobs (``ovlp_blob_energy``), and voxel size (``vox_size_x``, ``vox_size_y``, ``vox_size_z``). Each row corresponds to a different track, specified among the others within an event with its ``trackID``.
+ * ``/Summary/Events``: global information related to the event. Each row is one event.
+ * ``/DST/Events``: copy of the point-like information (**kdst**) of events, output of :doc:`penthesilea`.
+ * ``/Filters/hits_select``: flag to indicate if an event pass the selection of having more than 0 hits. 
+ * ``/Filters/topology_select``: flag to indicate if an event pass the selection of having less hits than the ``max_num_hits`` parameter specified in the config file.
```
tcontrer: event pass -> event passes

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+   * - ``energy_threshold``
+     - ``float``
+     - If the energy of one of the original end-point voxels is smaller than this value,
+
+       the voxel will be dropped and its energy redistributed to the neighbours.
+
+   * - ``min_voxels``
+     - ``int``
+     - The voxel dropping procedure commented on ``energy_threshold`` can only happen if the number of voxels is larger than the value specified in this argument.
```
tcontrer: on energy_threshold -> in energy threshold

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+       Events with more hits are filtered out and their ID will not appear in the ``Filters/topology_select`` table of the output.
+
+
+.. _Isaura workflow:
+
+Workflow
+--------
+
+The basic idea behind Isaura is to run the "traditional" *Paolina* algorithm over the deconvoluted hits output in :doc:`beersheba`, in order to convert ``events`` into ``tracks`` with ``blobs``. This process can be divided into two main steps:
```
tcontrer, line 82: Is there a doc we can reference for Paolina?

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+
+First of all, it is mandatory to perform some selections concerning the number of hits that events contain, in order to be able to compute all the tracking information for each of them.
+  
+The first condition that all events must fulfill to be processed is to contain at least one hit. If that does not happen, the event will be rejected, which will be displayed in the table ``Filters/hits_select``.
+
+The next step within the algorithm consists in checking that the number of hits is lower than the value provided in the config file (``max_num_hits``). That argument was introduced because, when running *Paolina* algorithm after :doc:`penthesilea`, there were some events that comprise such large amount of hits that the tracking information extraction took ridiculously long. The following picture shows the number of *Penthesilea hits* (**hdst**) per event (with a different scale) for a typical 24h-long low-background run included in the NEXT-White double-beta analysis [#]_. High energy (trigger2) events usually contain around 200 Penthesilea hits (as right panel points out), while there are some of them with more than 10000 hits (illustrated in left panel).
```
tcontrer: extraction took ridiculously long -> extraction took a ridiculously long time

tcontrer: while there are some of them with -> while there are some with

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+ .. image:: images/isaura/nhits_per_evt_r8571.jpg
+   :width: 1000
+
+The plot also shows that these events only  appear a few times within a 24h-long low-background run (around 0.05% of the total set of events). Rejecting this type of events is not a particularly worrysome issue: they would be thrown away in the posterior analysis, owing to the fact that none of them are exclusively contained inside the fiducial volume. The spatial distributions for one of these events is presented below (where the grey dashed lines illustrate the boundaries of the chamber).
```
tcontrer: Rejecting this type of events -> Rejecting this type of event

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+The following subsections explain each of these processes in detail.
+
+
+.. _Connectivity:
+ 
+**Separating events into tracks**
+
+
+Once events are properly selected according to the :ref:`previous subsection <Hits-based selection>`, their hits are grouped into 3D volume elements (``voxels``) with the objective of studying the connectivity. The size of these voxels is more or less fixed (depending on the ``strict_vox_size`` parameter in the config file), and their energy correspond to the sum of the energy of the hits included in the voxel. Following a Breadth-First Search (BSF) [#]_ algorithm, the voxels sharing side, edge, or corner will be part of the same **track**. The figure below shows the voxelization result of a real NEXT-White data (Run-VI) single-electron candidate of 1.73 MeV. In this case, after grouping the *deconvoluted hits* into [5 mm x 5 mm x 5 mm] voxels, the event was classified as single-track.
```
tcontrer: voxel sharing side -> voxels sharing a side

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+To compute the position of the blobs, we need to find the two extreme voxels of the track, which is done following the BFS algorithm. Then, as the figure below illustrates, the energy-weighted averaged position of the hits (represented with red stars) inside these voxels will correspond to the **blob center** (represented with the black dot from where the grey arrow starts).
+
+
+.. image:: images/isaura/blobs_position_definition.png
+   :width: 400
+   :align: center
+
+
+In consequence, the first thing to do is to localize the two end voxels for each track. Defining the distance between any pair of voxels as the shortest path **along the track** that connects them, the two extreme voxels will be the ones with the longest distance between them. However, there are two special cases that are important to comment:
```
tcontrer: In consequence, the first thing -> The first thing

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+ - Another particular scenario is the one that comes up when there are multiple end-voxel candidates (one can imagine that the shorter the track the more probable is this to happen). To deal with it, the more energetic candidates will be the ones set as extremes. With this convention, we aim to minimize the use of the voxel-dropping algorithm commented above for those cases where the energy of one candidate is larger than ``energy_threshold`` while the other one is below that value.
```
tcontrer: the more probable is this to happen -> the more probable this is to happen

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+
+From these blob centers, 3D spheres of radius ``blob_radius`` (specified in the config file) are taken. The hits inside the sphere will contribute to the energy of the blob, that will be stored as ``eblob1`` and ``eblob2`` [#]_. It is relevant to take into account here that not every hit falling inside the blob sphere will be considered for its total energy, but only the ones that belong to a voxel adjacent to the one labeled as extreme.
+
+
+
+The final step of the *Paolina* algorithm includes the computation of the ``ovlp_blob_energy`` (“*overlap blob energy*”) variable: in short tracks it is common to have **overlapping blobs**, i.e. blobs that share some of their hits [#]_. In these cases, the ``eblobi`` variables become meaningless, since the energy of these hits would contribute to both blobs [#]_. Therefore, it will be interesting to reject this type of events during the posterior analysis in case the blob energy distributions are intented to be exploited. Owing to the fact that the aforementioned variable is defined as the total amount of energy of these shared hits, a selection of ``ovlp_blob_energy = 0`` will get rid of the corresponding events easily.
```
tcontrer: reject this type of events -> reject this type of event

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+
+.. _Extracting the topology-related information:
+
+Extracting the topology-related information
+:::::::::::::::::::::::::::::::::::::::::::
+
+An excellent topological discrimination between signal and background (thanks to the usage of a gaseous medium inside the TPC) is one of the fundamental trademarks of the NEXT experiment. That is achievable thanks to the exploitation of the so-called **blobs**. They are defined as imaginary 3D spheres located around both ends of each track. Their energy is an excellent tool to investigate whether there has been a large and sudden energy deposition in the track extreme (i.e. *Bragg peak*, indicating the stopping point of a charged particle) or not (starting point of its trajectory). Therefore, that will be a crucial stage within the reconstruction chain, since if it is performed correctly, it will allow separating double-electron (such as the double-beta signal) from single-electron (the majority of backgrounds) tracks.
+
+
+In order to achieve that, it is necessary to:
```
halmamol, line 133: Maybe you can add again here the same links you used before

docs/source/isaura.rst
```diff
@@ -1,5 +1,209 @@
...
+ .. image:: images/isaura/XY_Z_distributions_evt_many_hits.jpg
+   :width: 1000
+
+In any case, one can easily infer from the plots that these events are not physical. On the contrary, they seem to correspond to either some kind of flash occurring in the chamber (like a mini-spark) or some fail in the electronics (after the saturation of an alpha particle, for example). The ID of the events that are removed from the reconstruction chain because of this reason will be specified in the table ``Filters/topology_select``, in order to keep track of this information.
+
+..
+ Next step includes another (quite obvious) check: at least one hit inside the event must have a well-defined energy. If not, the event will be also rejected, since no topological information could be extracted.
+
+
+Finally, every event is also required to contain hits with well-defined energy. For instance, events with all hits outside the krypton correction map boundaries will be thrown away, since their energy cannot be corrected and their ``Ec`` variable (*corrected energy*) will be ``NaN``.
```
halmamol: Maybe mention here where is Ec included?

docs/source/isaura.rst
ausonandres: No... Or at least, not yet.

halmamol (commented):
The comments i left before are already included. Just minor things to add and it would be fine from my side!

docs/source/isaura.rst
ausonandres: Sorry, do you mean in which table is `Ec`? or at what stage is this parameter computed?

ausonandres: I've rephrased the whole paragraph, trying to be clearer... Do you think it is better now?

docs/source/isaura.rst
halmamol: I'm not fully sure how to proceed with this, in fact. Paolina was never a city, but a set of functions. I'll talk with the rest of the team to see how to proceed here.

halmamol (approved):
This PR adds Isaura documentation on the City lists. It includes a detailed description of the city, and examples of each step. I approve it, and together with the main revision of @paolafer (and English check of @tcontrer), it can be merged into the main repository. Very good job @ausonandres!

PR #7, opened 2022-07-20
Add Esmeralda documentation

ausonandres:
This PR adds the documentation for the city Esmeralda.

You can see the latest version here: https://pr007-sw-docs.readthedocs.io/en/latest/esmeralda.html


docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+------
+
+ * ``/CHITS/highTh``: energy-corrected hits table of the ones that pass the higher charge theshold selection.
+ * ``/CHITS/lowTh``:  energy-corrected hits table of the ones that pass the lower charge theshold selection.
+ * ``/Tracking/Tracks``: topological information of events. Each row corresponds to a different track, specified with the ``trackID`` among the other possible tracks within an event.
+ * ``/Summary/Events``: global information related to the event. Each row is one event.
+ * ``/DST/Events``: copy of the point-like information (**kdst**) events, which constitutes one of the output tables of :doc:`penthesilea`.
+ * ``/Filters/high_th_select``: flag to indicate if an event passes the selection of having at least one hit with more charge than the ``threshold_charge_high`` parameter of the config file. 
+ * ``/Filters/low_th_select``: flag to indicate if an event passes the selection of having at least one hit with more charge than ``threshold_charge_low``. 
+ * ``/Filters/topology_select``: flag to indicate if an event passes the selection of having less (high-threshold) hits than ``max_num_hits``.
```
bpalmeiro: I guess the flags are also a list with a flag for each event. It may be worth explicating this.

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+   * - ``map_fname``
+     - ``string``
+     - Name of the krypton correction map to use.
+
+   * - ``threshold_charge_low``
+     - ``float``
+     - Lower threshold (in ``pes``) applied to the charge of hits.
+
+       The energy of the non-selected hits is transferred to the closest one (that passes the selection). 
+
```
bpalmeiro: This is more a procedural comment rather than what the parameter means. I'd consider removing it since it's explained in detail later.

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+     - ``bool``
+     - If *True*, temporal corrections are applied during the energy-correction procedure.
+
+       It must be set to *False* if:
+
+       (1) no temporal correction dataframe exists in map file
+
+       (2) the input is MC (events are not provided with timestamp variable).
+
+- On the other hand, ``paolina_params`` handle the processes associated to step (iii):
```
bpalmeiro: handles

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+   * - ``vox_size``
+     - ``[float,float,float]``
+     - X, Y, and Z dimensions of the voxels used in the voxelization of the hits.
+
+   * - ``strict_vox_size``
+     - ``bool``
+     - Flag to indicate if the size of the voxels is forced to be exactly the values provided in the previous argument (*True*),
+
+       or, on the other hand, if they are allowed to change a bit for each track, aiming to optimize the voxelization process (*False*).
```
bpalmeiro: Just picky, I know. But for clarity, I'd rephrase this to something like:
If True: the voxels size is forced to be exactly the values provided in the previous argument,
If False: the voxel dimensions are allowed to change a bit for each track, aiming to optimize the voxelization process.

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+       the voxel is dropped and its energy redistributed to the neighbours.
+
+   * - ``min_voxels``
+     - ``int``
+     - The voxel dropping procedure commented on ``energy_threshold`` can only happen if the number of voxels is larger than the value specified in this argument.
+
+   * - ``blob_radius``
+     - ``float``
+     - Radius of the blobs.
```
bpalmeiro: It may be worth mentioning where this radius is measured from. Voxel centre?

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+       Events with more hits are filtered out and their ID will not appear in the ``Filters/topology_select`` table of the output.
+
+
+.. _Esmeralda workflow:
+
+Workflow
+--------
+
+Since its creation, *Esmeralda* has been a sort of a *Frankenstein*'s monster, seeing that it contains two main parts of code that are uncorrelated. They are planned to be decoupled at some point, but in the meantime both of them will be performed here:
```
bpalmeiro: meantime,

are performed here?

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+ #. :ref:`Manipulation of the SiPM-based hits <Manipulation of SiPM-based hits>`
+
+    * :ref:`Energy reassignment <Energy reassignment>`
+    * :ref:`Energy calibration <Energy calibration>` 
+
+ #. :ref:`Topology information extraction of events <Topology information extraction>`
```
bpalmeiro, line 144: Here you use other criteria for labelling the procedure that I kind of prefer since it stresses the two different tasks. Anyway, I'd choose between I, II, and III or Ia, Ib, and II.

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
 Esmeralda
-==========
+=========
+
+*From Spanish esmeralda (“emerald”), as first used in the novel Notre-Dame de Paris (1831) by Victor Hugo.*
+
+At the moment of writing this, the city `Esmeralda` has three main purposes:
+
+ **(i)** to apply a cut to remove the *SiPM hits* with charge below a certain threshold,
```
bpalmeiro: a charge

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+ #. :ref:`Topology information extraction of events <Topology information extraction>`
+
+
+.. _Manipulation of SiPM-based hits:
+
+Manipulation of the SiPM-based hits
+:::::::::::::::::::::::::::::::::::
+
+
+The **hits** (also known as *clusters*) that constitute the input for this city are the ones coming from :doc:`penthesilea` --inside the table ``RECO/Events``--. They are energy depositions separated in time slices  [#]_ of 2 :math:`\mu s` provided with a given *X* and *Y* position in the chamber, according to the pattern of signals observed on the SiPMs. Therefore, these energy depositions combine the light collected by both the SiPMs (tipically called "*charge*" and stored in the dst as ``Q``) and the PMTs ("*energy*" --``E`` variable-- for us) [#]_. Due to the fact that PMTs sentitivity is better than the one of SiPMs, there might be some time slices in the PMTs waveform that do not appear in the SiPMs one. In these cases, a hit is created at the correponding position along the DT-axis, with charge set to ``NaN`` and  ``X = Y = 0``.
```
bpalmeiro: typically
sensitivity
corresponding
with a charge

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+.. _Energy reassignment:
+
+**Reassignment of the hits energy**
+
+The main features of the input dst, as well as how events might contain hits with non-defined (``NaN``) charge --which leads to non-defined position at the *XY* plane-- has been commented above. As a consequence of this last fact, the first thing to do consists in applying a cut on the charge of hits in order to deal with this issue. Besides that, this selection also removes the hits with very low charge, aiming to obtain a "cleaner" version of the event, what will allow us to perform a better reconstruction in the posterior analysis. Therefore, if one hit does not pass the charge threshold (``threshold_charge_low`` or ``threshold_charge_high``, depending on the case), its energy is charged-weighted redistributed between those ones that do pass the cut and belong to the same time slice. After that, the hit will be removed from the dst. In case none of the hits for a given slice is above the threshold, a new hit containing all the redistributed energy is created at the same *Z* coordinate position, with: ``Q = NaN`` and ``X = Y = 0``.
```
bpalmeiro: with a non-defined charge [...] a non-defined position
have been commented
consists of
a very low charge
event, which will allow us
is charged-weighted and redistributed

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+At this point, the dst may include time slices with undefined charge (although defined energy) as a result of the previous step. This issue is now addressed by redistributing their energy among the closest hits along the *Z*-axis, and that comprise the same ``npeak`` (S2 peak) in case  ``same_peak`` is set to *True*. This energy sharing is proportional to the energy of the "good" hits. If all hits within a peak are ``NaN``, the S2 would be reconstructed as being empty.
```
bpalmeiro: slices with an undefined charge

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+
+
+
+
+.. _Energy calibration:
+
+**Hit energy calibration**
```
bpalmeiro: To be consistent: calibration of the hit energy?

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+Once all the hits for the analysis has been selected and are well-defined (with a proper ``E``, ``X``, ``Y`` and ``DT``), it is time to correct them using the krypton-based map generated in :doc:`ICAROS`. Its name will be specified with ``map_fname`` in the configuration file.
```
bpalmeiro: have been selected

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+Thanks to the **dual trigger scheme** exploited by the experiment, we are capable of taking low-energy krypton runs simultaneously to the usual high-energy data. This allows us to monitor the status of the chamber in every moment and to obtain a correction map specific for every run (providing a very robust tool, that avoids possible temporal discrepancies in case the krypton run and the physics run would have been taken at different moments).
```
bpalmeiro: status of the chamber at every moment

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+ - The **XY non-homogeneities** in the light collection by the PMTs. This time, the *geometry*  (``e0``) map will be applied.
+
+An example of these correction maps can be seen below (left: geometry map, right: lifetime map), for a NEXT-White high-energy calibration run [#]_ (namely run 8183, taken just before Run-VI). 
+
+.. image:: images/esmeralda/maps_r8183.png
+   :width: 900
+   :align: center
+
+
+In NEXT-White, the XY pattern distributions at both krypton maps were demonstrated to remain stable during the extensive data-taking period devoted to the double-beta analysis. Nevertheless, their mean value did variate in time, due to the improvement in the purity of the gas inside the chamber, for example, thanks to its recirculation. This temporal variations are exemplified in the plots below, where the evolution of lifetime, energy scale (e0) and drift-velocity are represented for the same high-energy calibration run as before.
```
bpalmeiro: variate over time?

This temporal variation or these temporal variations

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+In NEXT-White, the XY pattern distributions at both krypton maps were demonstrated to remain stable during the extensive data-taking period devoted to the double-beta analysis. Nevertheless, their mean value did variate in time, due to the improvement in the purity of the gas inside the chamber, for example, thanks to its recirculation. This temporal variations are exemplified in the plots below, where the evolution of lifetime, energy scale (e0) and drift-velocity are represented for the same high-energy calibration run as before. 
+
+.. image:: images/esmeralda/maps_temporal_evolution.png
+   :width: 900
+   :align: center
+
+If the ``apply_temp`` parameter is *True*, these variations will be considered for the correction. On the contrary, this variable must be set to *False* in case the map does not include this temporal information table, or the city is run over MC files (whose events do not have a timestamp variable either).
+
+
+The following image shows the clear improvement on the energy spectrum after applying all the corrections explained above:
```
bpalmeiro: improvement in the

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+
+
+
+
+|
+|
+
+.. warning::
+ Despite the energy correction explained just before, there are **two** important factors that make the previous calibration not ultimate. They are not going to be explained in detail here, seeing that these further corrections are not applied along `Esmeralda`. However, since they are not applied inside any other city either and the energy modification of events is performed here, it is justified to comment them now.
```
bpalmeiro: In addition to the energy correction...? 

comment on them

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+
+
+
+|
+|
+
+.. warning::
+ Despite the energy correction explained just before, there are **two** important factors that make the previous calibration not ultimate. They are not going to be explained in detail here, seeing that these further corrections are not applied along `Esmeralda`. However, since they are not applied inside any other city either and the energy modification of events is performed here, it is justified to comment them now.              
+
```
bpalmeiro: as a general comment for this section, I think it is not important to have it right here since it's not a city-related issue. I'm fine with the explanation for completeness, but I'd suggest moving the discussion to some sort of appendix.

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+1. **Non-linearities at high energies**. Due to the significant difference between the krypton energy scale and the one of the physics data (above 1 MeV), the Kr-based energy correction might **not be sufficient** for all the energy range considered. Therefore, although krypton maps were applied, it is advisable to check the high energy peaks, so as to account for observed **non-linearities** and obtain the proper calibration. The plots below show clearly how the high energy 208-Thallium gamma lines (nominal values are illustrated with dashed red lines) are not aligned perfectly in spite of the maps corrections.
+
+.. image:: images/esmeralda/energy_spectrum_corr_vs_uncorr_PEAKS.png
+   :width: 900
+   :align: center
+
+*Note:* the energy scale of the uncorrected hits coming from the hdst presented in the plots above has only illustrative pruposes (to compare both distributions). Its conversion between pes to MeV was made considering in an approximate way the number of pes yielded by krypton events.
```
bpalmeiro: purposes

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+
+2. **The axial length (Z-width) effect**. There is an additional final energy correction that must be applied in the analysis post-reconstruction: the so-colled *Z-width effect* correction. Detailed information about possible explanations for this phenomenon, as well as an empirical procedure to deal with it can be found `here <https://arxiv.org/abs/1905.13110>`_.
```
bpalmeiro: so-called

the link may be to the actual published paper and not to the axiv, maybe to the inspirehep entry?

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+
+
+Apart from the energy correction, the position of hits along the drift time (``DT`` variable in the dst) of the chamber is also transformed to its equivalent in the **Z-axis** [#]_. To do that, drift time values are simply multiplied by the *drift-velocity*. In data, this magnitude is computed as a funtion of time, so this conversion can be also time-dependent if ``apply_temp`` is *True*.
```
bpalmeiro: drift velocity (no -)
a function of time
also be

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+
+.. note::
+ At this point, and once all the correction process (made inside `Esmeralda`) is explained, it is relevant to remark that there are several possible high-level analysis to be conducted posterior to this city. Each of them require a different treatment of its input, being that the reason why all the processes commented before are performed **two** times for all events every time the city is run.
```
bpalmeiro: analyses
requires

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+ - The current *official* reconstruction (that will run :doc:`beersheba` after this) and the DNN analysis are interested in keeping also some lower charged hits (all the ones above **5-10 pes**, typically). This is due to the fact that both analysis will carry out additional manipulations to the data, and in consequence, they ought to keep more information of events. In these cases, the lifetime inside `Esmeralda` of the energy corrected hits that passed the ``threshold_charge_low`` threshold will end here. They will be stored in the ``CHITS/lowTh`` table of the output, as indicated in ``Filters/low_th_select``.
```
bpalmeiro: analyses (plural) in general
also keeping
more information about events
energy-corrected hits

docs/source/esmeralda.rst
```diff
@@ -1,5 +1,277 @@
...
+
+ - On the other hand, the "classical" [#]_ analysis performs all the tracking algorithms directly over these "high-pitched" (in comparison to the deconvoluted ones, out of :doc:`beersheba`) SiPMs hits. In this case, the :ref:`Paolina <Topology information extraction>` algorithm, which is described as follows, will correpond to the latest step of the data processing for them. It was demonstrated that in order to obtain a clearer track and perform a more accurate tracking reconstruction, a sharper (around **30-35** pes, specified in ``threshold_charge_high``) charge threshold cut is desired. The resulting hits will appear in the table ``CHITS/hightTh``, according to ``Filters/high_th_select``.
```
bpalmeiro: correspond
CHITS/highTh (extra t)

bpalmeiro (commented):
As a first round, here you have some comments, mainly about the language, style, and structure. It is indeed a Frankenstein's monster, and there's a lot of information to digest, let me do a second review more centred on how the information is delivered, but for now... here you have. Impressive work!

docs/source/esmeralda.rst
ausonandres: These Filters tables correspond to two-column tables that comprise the event ID (``event``) and the flag (``passed``), so it's quite straightforward, do you think it is still worth commenting?

docs/source/esmeralda.rst
ausonandres: Ok, sounds good.

docs/source/esmeralda.rst
ausonandres: Ok, good point!

jahernando:
Hi Alberto, sorry for the late comments, if you want we can chat privately.

I will make a comment per section to keep track

- Esmeralda
	- _At the moment of writing this, the city Esmeralda has two main purposes:_ (remove, not necessary). The city of Esmeralda has two purposes:
	- _A first (and rough) computation of the topology-related information of events (analogous to [Isaura]_. (remove, not necessary)
	- I do not understand, remove or rephrase, maybe not needed, stages I) and ii) are in fact 1) and 2): _It is worth remarking here –in order to understand [Output] and [Config]– that stages (i) and (ii) are performed two times inside the city for two different subsamples of the input dst. Thanks to that, the output is adapted to two different possible posterior analyses. This will be further discussed later on, in the [Workflow]section._

jahernando:
- Input:
	-  comment briefly about what is there on the input tables

jahernando:
- Output:
	- Indicate better the output of the main tables: [/CHITS/highTh],[ /CHITS/lowTh], [/Tracking/Tracks], [/Summary/Event], similar to what is done in Isaura documentation

jahernando:
- Configuration
	- _icor_hits_params_  is in charge of the information related to steps (i) and (ii) commented in the introduction. (you mean 1.a) and 1.b)
	- _map_fname_: Indicate that the KrMap contains the geometrical corrections computed to this specific run number using the Kr-calibration data (either real or simulated data)
	- indicate that "charge" is the light measured by the SiPMs. ("charge" is NEXT's jargon)
	- _same_peak_: I do not understand it. You explain it latter, but indicate here that sometimes in a time slice of the wf there is PMTs signal but not SiPMs signal, in this case a 'fake' hit is created with 'Nan' "charge" and position X=0, y= 0, is this correct?
	- _apply_temp_: I understand temporal corrections refer to the evolution of the lifetime corrections, right? Indicate so
	- On the other hand, paolina_params handles the processes associated with step (iii), (iii-> 2)
	- _energy_threshold_: What does mean "end-point" voxel?
	- _min_voxels_: rephrase it. The minimum number of voxels required to apply the Paulina algorithm?
	- _max_num_hits_: An explanation of why this cut is applied (computing time?), what is its default value, and why? This cut can be dangerous, it can remove some relevant physical events (maybe muons, alphas...)

jahernando:
- Workflow:
	- Remove: _Since its creation, Esmeralda has been a sort of a Frankenstein’s monster, seeing that it contains two main parts of code that are uncorrelated. They are planned to be decoupled at some point, but in the meantime, both of them are performed here._
	-  And replace it. Something like this: "In its current configuration, Esmeralda runs two algorithms: selects hits in two groups depending if they pass a low or a high energy threshold on the SiPM signal and calibrate them using the Kr calibration map; and then runs the topology algorithm "Paulina" which first creates 3D voxels of a given size, using the high-threshold passing hits, and then select the maximum of the minimum spanning tree connecting any two voxels; finally the extremes of this tree serve as seeds to compute the two blobs of the track"

jahernando:
- Manipulation of the SiPM-based hits
	- This part: _Apart from that, it is relevant to remind that the energy of the input hits cannot be used directly in the high-level analysis because of two reasons: It is stored according to the pes scale (thanks to the ADC-to-pes conversion of the PMT waveforms performed in [Irene]. It must be corrected due to different processes that degrade the light collection. With all the information presented above, one could realize that the SiPM-based hits that enter the city must suffer some modifications in order to be useful for the later analysis. The explanation of these processes is the main purpose of this section._ Needs to be more clear, rephrase it: The "E" of the hits is in pes and needs to be converted to "eV" using the geometrical and lifetime correction provided in the KrMap computed for this run using the calibration Kr data.
	- This part is not necessary: _The main features of the input dst, as well as how events might contain hits with a non-defined (NaN) charge –which leads to a non-defined position at the XY plane– have been commented on above. As a consequence of this last fact, the first thing to do consists in applying a cut on the charge of hits in order to deal with this issue. Besides that, this selection also removes the hits with a very low charge, aiming to obtain a “cleaner” version of the event, which will allow us to perform a better reconstruction in the posterior analysis. Therefore,_. The rest of the text here is ok
	- _This energy sharing is proportional to the energy of the “good” hits._ I do not understant this! rephrase!

jahernando:
- Calibration of the hits energy
	- Less colloquial here: _Once all the hits for the analysis have been selected and are well-defined (with a proper E, X, Y, and DT), it is time to correct them_. Something like this: The selected hits, whose have (x, y, dt, E) components, are calibrated using the Kr-Map obtained form Kr Data to convert the energy, E, from pes into eV, and the drift time, DT, in us, into z position, in mm. Now indicate the two corrections on lifetime/attatchment , ii) geometrical correction, ii) add coversion of dt into z using the drift-velocity also computed in the map.   
	- Again, less colloquial here: _In addition to the energy correction explained just before, there are two important factors that make the previous calibration not ultimate. They are not going to be explained in detail here, seeing that these further corrections are not applied along Esmeralda. However, since they are not applied inside any other city either and the energy modification of events is performed here, it is justified to comment on them now._ Something like this: In addition to the energy correction explained above there are two other energy corrections that need to be taken into account: I) the non- linearity of the energy scale (show the plots) and ii) the energy dependence of the S2 peak on its time - length (usually called "delta-z effect").
	- This should have been explained before! _Apart from the energy correction, the position of hits along the drift time (DT variable in the dst) of the chamber is also transformed to its equivalent in the Z-axis [4](https://pr007-sw-docs.readthedocs.io/en/latest/esmeralda.html#id10). To do that, drift time values are simply multiplied by the drift velocity. In data, this magnitude is computed as a function of time, so this conversion can also be time-dependent if apply_temp is True._
	- All the text in the Note it too verbose. Just simply it. Something like this: Two sets of calibrated hits are saved in the output: I) in [/CHITS/lowTh] are the hits that pass a low threshold (>5 pes typically) on the signal SiPMs and are later used in other cities (Beersheba for example) to perform a more fine reconstruction. ii) in [/CHITS/highTh] are the hits that pass a high threshold ( > 30 pes typically) on the single of the SiPMS and are used by the Paulina algorithm that follows.

jahernando:
- Topology information extraction
	- The text in this section is not descriptive enough. Either you locate here the comments that you have put in the  "Isaura" documentation, or you can refer to them, and leave here something like this: The second step of Esmeralda runs the Paulina reconstruction algorithm. The algorithm has the following steps:
		- High Threshold hits are converted into voxels of sizes given by [vox_size ] typically (15 x 15 x 15 mm3) 
		- A graph module is used to compute the maximum of the minimum spanning trees that can be computed with any pair of voxels.
		- blobs are computed starting from the extremes of that tree. The energy of the blob is the energy of the hits inside a sphere of radius [blob_radius], which center is computed from the extremes of the selected tree.
	- Now you need to explain the two tables: [/Tracking/Tracks] and [/Summary/Events] that output Esmeralda. 
	- You can comment that the "Paulina" algorithm also runs in "Isidora" city, but there it uses and input the deconvoluted hits, which are the output of Beerseba's city; and that output the same tables of tracks and Summary.
	- The 3D plots of the hits and voxels are quite useful, If you can add an example with the 'blobs' that will be great!

ausonandres:
@jahernando, I've just moved your comments to the file itself, to make things easier...

ausonandres:
Concerning the `input`, I was just following the model we had (https://next-exp-sw.readthedocs.io/en/latest/irene.html), where there aren't any explanations for the inputs. I could add them, but I guess they should be added to every city as well. Maybe @halmamol can say something about this.

halmamol:
> Concerning the `input`, I was just following the model we had (https://next-exp-sw.readthedocs.io/en/latest/irene.html), where there aren't any explanations for the inputs. I could add them, but I guess they should be added to every city as well. Maybe @halmamol can say something about this.

@jahernando So far we haven't included any detailed description about the `input` This should be described as the output from the previous city. What we have done for cities like Irene or Buffy, that retrieve files directly from the detector or simulations is to specify along the workflow the type of tables we use as input. Maybe this is something we should discuss in our software meetings.

jahernando:
> > Concerning the `input`, I was just following the model we had (https://next-exp-sw.readthedocs.io/en/latest/irene.html), where there aren't any explanations for the inputs. I could add them, but I guess they should be added to every city as well. Maybe @halmamol can say something about this.
> 
> @jahernando So far we haven't included any detailed description about the `input` This should be described as the output from the previous city. What we have done for cities like Irene or Buffy, that retrieve files directly from the detector or simulations is to specify along the workflow the type of tables we use as input. Maybe this is something we should discuss in our software meetings.
> 

Ok, if this is the usual procedure. It is ok to not have comments about the tables in the Input

jahernando (approved):
The correction and extension of the comments improved the understanding to the documentation.

bpalmeiro (approved):
This PR adds detailed documentation of Esmeralda. The explanation is clear and easy to follow, as well as complete regarding the reconstruction key features. Good job!

PR #9, opened 2022-09-05
Add Buffy documentation

halmamol:
This PR includes Buffy documentation in the readthedocs.

It can be checked here: https://pr009-sw-docs.readthedocs.io/en/latest/buffy.html


docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
 Buffy
 ==========
 
-.. caution::
-  This page is currently not created, we are working on it. If you would like to contribute on this documentation, reach `me <helena.almamol@gmail.com>`_!
+*sorts MC sensors info into buffers, slays vampires, etc*
```
gonzaponte:
```suggestion
*Sorts MC sensors info into buffers, slays vampires, etc*
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+
+Full simulations generate sensors (PMTs and SiPMs) response,
+including the time and the detected charge of a specific signal. However, this sensor
+information does not have the waveform format obtained from data taking.
```
gonzaponte: from the detector maybe better?

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+*Buffy* takes nexus sensors information, and sorts them into *true waveforms* (**TWF**).
+**TWF**\ s represent the signal amplitude of a given sensor without any type of
+distortion or effect, within a certain time interval and given the sensor sampling time.
```
gonzaponte: effect from the electronics, to be more specific.

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+This type of time ordering of the sensor signal in a data-like format is what is called *bufferisation*.
+**TWF**\ s generated from *Buffy* can be transformed into *raw waveforms* **RWF**\ s with :doc:`diomira`.
```
gonzaponte: into raw waveforms (i.e. with effects from the electronics)

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+For detector geometries without sensor electronics, **TWF**\ s can be transformed directly into **PMap**\ s
```
gonzaponte: This is not 100% accurate. While `Diomira` uses a realistic model for the PMT electronics, `Hypathia` uses a simplified one (gaussian fluctuations around the baseline +  charge fluctuation for pes).  Both simulate the electronics, but the latter doesn't save the resulting waveforms and feeds them into the pmap creation algorithms.

Perhaps "incomplete electronics modeling" is a better wording, although it is also more obscure...

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+ * ``/MC/sns_response``
+
+.. _Buffy output:
+
+Output
+------
+
+ * ``/Run/runInfo``: run info table
+ * ``/Run/events``: event info table
+ * ``/Run/eventMap``: event and nexus event for each index
```
gonzaponte: what does "for each index" mean?

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+ * ``/RD/pmtrd/``: time ordered signal amplitude of the PMTs in true photoelectrons (PMT buffers). array: number of events, number of PMTs, length of PMT waveform (:math:`\mu`\ s).
```
gonzaponte: array with shape: (bla, bla, bla).

Also is the length in mus? I thought it was in units of the sampling period.

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+   :widths: 50 40 120
+   :header-rows: 1
+
+   * - **Parameter**
+     - **Type**
+     - **Description**
+
+   * - ``max_time``
+     - ``int``
+     - Maximal length of the event that will be taken into account starting from the first detected signal. All signals after that are lost. Must be greater than ``buffer_length``. If not, raises a warning and set ``max_time`` == ``buffer_length``
```
gonzaponte:
```suggestion
     - Maximum duration of the event that will be taken into account starting from the first detected signal. All signals after that are lost. Must be greater than ``buffer_length``. If not, raises a warning and set ``max_time`` == ``buffer_length``
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+.. list-table::
+   :widths: 50 40 120
+   :header-rows: 1
+
+   * - **Parameter**
+     - **Type**
+     - **Description**
+
+   * - ``max_time``
+     - ``int``
```
gonzaponte: is it really an int? It doesn't sound right unless it is in units of the sampling period, but I thought it was in mus. This should also be specified.

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+   * - ``trigger_threshold``
+     - ``int``
+     - Trigger threshold for selection in pes.
+
+
+.. _Buffy workflow:
+
+Workflow
+--------
+For full simulations, NEXUS ``/MC/sns_response`` table stores for each event (``event_id``) the list of sensors (``sensor_id``) that detect a specific number of photons (``charge``) in an ordered number of the time bin (``time_bin``).
```
gonzaponte, line 74: The fact that the data from nexus comes binned in time is quite important. It's worth writing it either here as a footnote or somewhere in the introduction. Maybe we should also link to the part of the nexus documentation that specifies how to set this binning (if it exists...)

I've seen now that this is briefly mentioned in the histogram creation section. I think that we should clarify earlier that the binning comes from nexus.

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+
+
+This process is separated in the following tasks in the city:
+
+• :ref:`Histogram creation <Histogram>`
+• :ref:`Signal Search <Signal-Search>`
+• :ref:`Synchronisation and trigger separation <Trigg-Separation>`
+
+.. note::
+  Historically, Buffy is based in an initial code of detsim (https://github.com/next-exp/IC/tree/master/invisible_cities/detsim) and most of its functions are located in that path but they are independent to :doc:`detsim` city.
```
gonzaponte, line 98: Uff... terrible code organization :rofl:

gonzaponte, line 98:
```suggestion
  Historically, Buffy is based on an initial code of detsim (https://github.com/next-exp/IC/tree/master/invisible_cities/detsim), and most of its functions are located in that path, but they are independent of the :doc:`detsim` city.
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+
+Histogram creation
+::::::::::::::::::
+
+Takes NEXUS information about sensor hits (``/MC/sns_response``). Checks time stamp of an event according to the sensors response, and defines a histogram between [:math:`t_{min}`, :math:`t_{max}`], being:
+
+• :math:`t_{min}`: the time stamp of the first charge deposition of the event,
+• :math:`t_{max}`: defined considering that ``max_time`` =  :math:`t_{max}` - :math:`t_{min}`.
+
+Once these histograms are defined, they are sampled according to the binning of each sensor (``pmt_width`` and ``sipm_width``). Sampling widths are included in the simulation parameters (``/MC/info``), and depends on the type of sensor and detector. Normally corresponds to 25 :math:`\mu`\ s for PMTs and and 1 :math:`\mu`\ s for SiPMs.
```
gonzaponte:
```suggestion
Once these histograms are defined, they are sampled according to the binning of each sensor (``pmt_width`` and ``sipm_width``). Sampling widths are included in the simulation parameters (``/MC/info``), and depends on the type of sensor and detector. Normally corresponds to 25 :math:ns for PMTs and and 1 :math:`\mu`\ s for SiPMs.
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+
+.. _Signal-Search:
+
+Signal Search
+::::::::::::::::::
+
+Charge is distributed in previously defined histograms. The code searches for signal-like charge according to a given threshold (``trigger_threshold``). Simple threshold value is applied on binned charge. Once it is found, it defines a trigger time, :math:`t_{trigger}`. PMT/SiPM sum is ordered by the given buffer length (``buffer_length``) considering :math:`t_{trigger}` and the ``pre_trigger`` configuration. Waveforms are therefore defined for each sensor in a specific length based on ``buffer_length``/``sensor_width``.
```
gonzaponte: I find this paragraph hard to read, and I think it can be simplified. The relevant bit is that we look for the first value of the binned charge above a certain threshold, which defines our t_trigger. Then, I don't understand what is happening with the PMT/SiPM sum. It's ordered [...] considering [...], which is a touch too generic.

Also, we should mention earlier in the paragraph that this search for charge over threshold happens on the PMT sum.

docs/source/buffy.rst
```diff
@@ -1,5 +1,128 @@
...
+
+.. note::
+  :math:`t_{min}` does not need to be at 0, since it is defined based on the first charge deposition, independently if it is above the ``trigger_threshold`` or not.
+
+.. _Trigg-Separation:
+
+Synchronisation and trigger separation
+:::::::::::::::::::::::::::::::::::::::
+
+Since the ``sensor_width`` is different for each sensor, it is necessary to align and synchronises the clocks between SiPMs and PMTs. Waveforms are sliced then according to binning (``pmt_width`` and ``sipm_width``), trigger time and configured pre-trigger (``pre_trigger``). In addition, it pads with zeros where it is necessary: arrays with zeros where there is no recorder pes in nexus (in the other sensors). Arrays with zeros where no signal and nexus recorded pes otherwise. If more than one trigger is found separated from each other by more than a buffer width, the nexus event can be split into multiple data-like triggers.
```
gonzaponte:
```suggestion
Since the ``sensor_width`` is different for each sensor, it is necessary to align and synchronise the clocks between SiPMs and PMTs. Waveforms are sliced then according to binning (``pmt_width`` and ``sipm_width``), trigger time and configured pre-trigger (``pre_trigger``). In addition, it pads with zeros where it is necessary: arrays with zeros where there is no recorder pes in nexus (in the other sensors). Arrays with zeros where no signal and nexus recorded pes otherwise. If more than one trigger is found separated from each other by more than a buffer width, the nexus event can be split into multiple data-like triggers.
```

gonzaponte: I think it is important to clarify that we shift the times of the charge histogram such that the first value over threshold falls at t = pre_trigger.

I would rephrase the rest omitting the padding part. It is probably clearer to say that the waveform is an array of zeros filled with the charge from nexus at the corresponding times. The padding is probably an implementation detail that doesn't help the casual reader.

gonzaponte (commented):
The images help a lot to visualize the process, good job!

docs/source/buffy.rst
halmamol: What do you think about something like this?
```suggestion
For detector geometries without sensor electronics, a simplified electronics modeling (gaussian fluctuations around baseline + charge fluctuations for pes) can be applied with :doc:`hypathia` to transform **TWF**\ s into **PMap**\ s.
```

docs/source/buffy.rst
halmamol: I've been crosschecking, and it's defined as `int` ([here](https://github.com/next-exp/IC/blob/666647c299b559163d47d9b4a17617edfea5886f/invisible_cities/cities/components.py#L991) and here [here](https://github.com/next-exp/IC/blob/666647c299b559163d47d9b4a17617edfea5886f/invisible_cities/detsim/buffer_functions.py#L73)).

docs/source/buffy.rst
gonzaponte: I now realize that the stuff in parentheses is only part of the story. I am not sure that those two steps are the complete simulation of the PMTs electronics. On top of that, we also ignore the simulation of the SiPMs electronics, which are surely part of `Hypathia`'s job. So perhaps it is better to leave it as "simplified electronics modeling" without the parentheses and refer to the `Hypathia` documentation.

I would also like to keep the word "directly" at the end to indicate that the intermediate waveforms are not stored :)

halmamol: Got it! We just need to be sure that it is emphasized on `Hypathia` documentation.

docs/source/buffy.rst
halmamol: It's not clear, i'll edit it.

docs/source/buffy.rst
halmamol: You're right! Changed, together with the array.

docs/source/buffy.rst
halmamol: I've added a link to the nexus wiki where the output format is especified and I've extended a little bit the explanation in the Histogram creation section.

docs/source/buffy.rst
halmamol: I've added this to clarify that this detsim folder it is not fully for detsim city. I hope it's clearly explained.

docs/source/buffy.rst
halmamol: I agree that it could be a little bit confusion, I've removed the info about it and rephrase a little bit earlier in the Histogram creation.

docs/source/buffy.rst
halmamol: I've rephase it this part a little bit to make it more clear!

docs/source/buffy.rst
gonzaponte: Yes, this was not a comment for you, just a statement for us as IC developers!

halmamol:
I have added the comments, and rephrased it a little bit some parts to make it more clear.

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
 Buffy
 ==========
 
-.. caution::
-  This page is currently not created, we are working on it. If you would like to contribute on this documentation, reach `me <helena.almamol@gmail.com>`_!
+*Sorts MC sensors info into buffers, slays vampires, etc*
+
+Full simulations generate sensors (PMTs and SiPMs) response,
```
gonzaponte:
```suggestion
Full simulations generate sensor (PMTs and SiPMs) responses,
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+including the time and the detected charge of a specific signal. However, this sensor
+information does not have the waveform format obtained from the detector.
+*Buffy* takes nexus sensors information, and sorts them into *true waveforms* (**TWF**).
```
gonzaponte:
```suggestion
*Buffy* takes nexus sensors' information and sorts it into *true waveforms* (**TWF**).
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+**TWF**\ s represent the signal amplitude of a given sensor without any type of
+distortion or effect from the electronics, within a certain time interval and given the sensor sampling time.
```
gonzaponte: did you mean this?
```suggestion
distortion or effect from the electronics within a certain time interval given by the sensor sampling time.
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+This type of time ordering of the sensor signal in a data-like format is what is called *bufferisation*.
+**TWF**\ s generated from *Buffy* can be transformed into *raw waveforms* **RWF**\ s (i.e. with effects
```
gonzaponte:
```suggestion
**TWF**\ s generated by *Buffy* can be transformed into *raw waveforms* **RWF**\ s (i.e. with effects
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+ * ``/MC/particles``
+ * ``/MC/sns_response``
+
+.. _Buffy output:
+
+Output
+------
+
+ * ``/Run/runInfo``: run info table
+ * ``/Run/events``: event info table
```
gonzaponte, line 33: Are these tables generated for the first time in the processing chain here? If so, the documentation should say it explicitly

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+
+Workflow
+--------
+For full simulations, NEXUS ``/MC/sns_response`` table stores for each event (``event_id``) the list of sensors (``sensor_id``) that detect a specific number of photons (``charge``) in an ordered number of the time bin (``time_bin``).
+
+ .. image:: images/buffy/MC_sns_response_table.png
+   :width: 300
+
+More details about nexus output can be found in its `Github Wiki <https://github.com/next-exp/nexus/wiki/Output-format>`_ . This type of tables do not have the same shape that the waveforms collected when data taken,
+and in addition they only provide the information from the sensors and time bins when some charge is detected. Buffy takes this information (``event_id``, ``sensor_id`` and ``time_bin``), and transforms it into the waveform
```
gonzaponte, line 80: > and in addition they only provide the information from the sensors and time bins when some charge is detected

Great, I think this sentence is crucial!

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+shape of the **TWF** expected for each type of sensor: ``pmtrd`` and ``sipmtrd`` (PMTs and SiPMs respectively) based on the *bufferisation* parameters provided.
+
+.. image:: images/buffy/pmt_wft.png
+  :width: 350
+.. image:: images/buffy/sipm_wft.png
+  :width: 350
+
```
gonzaponte, line 87: Is there a way to put captions on these images? I'm guessing these are examples of the city's output from the context, but it would be great if we could add a caption. Otherwise, we can just put a line above saying that these are some examples.

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+
+.. _Histogram:
+
+Histogram creation
+::::::::::::::::::
+
+As it was highlighted earlier, NEXUS information about sensor hits (``/MC/sns_response``) comes binned in time based on when a sensor sees some energy deposition.
+This means that ``time_bin`` column numbers are increasing for a given event, but they can have gaps, since the time bins where no charge is detected
+do not appear in the table. This initial part of the city, checks the time stamp of an event according to the sensors response, and defines histograms of charge distribution
```
gonzaponte: how about this?
```suggestion
This means that the ``time_bin`` column numbers are increasing for a given event, but they can have gaps since empty time bins are not stored. This initial part of the city checks the time stamp of an event according to the sensors' response and defines histograms of charge distribution
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+between [:math:`t_{min}`, :math:`t_{max}`], being:
+
+• :math:`t_{min}`: the time stamp of the first charge deposition of the event,
+• :math:`t_{max}`: defined considering that ``max_time`` =  :math:`t_{max}` - :math:`t_{min}`.
+
+.. image:: images/buffy/histogram_creation.png
+  :width: 800
+
+In this step, it also effectively pads with zeros in between separate signals. When these histograms (one for PMTs and another for SiPMs) are defined, they are sampled according to the binning of each sensor (``pmt_width`` and ``sipm_width``).
+Sampling widths are included in the simulation parameters (``/MC/info``), and depends on the type of sensor and detector. Normally corresponds to 25 :math:`ns` for PMTs and and 1 :math:`\mu s`\ s for SiPMs.
```
gonzaponte:
```suggestion
Sampling widths are included in the simulation parameters (``/MC/info``), and depends on the type of sensor and detector. Normally corresponds to 25 :math:`ns` for PMTs and and 1 :math:`\mu s`\  for SiPMs.
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+do not appear in the table. This initial part of the city, checks the time stamp of an event according to the sensors response, and defines histograms of charge distribution
+between [:math:`t_{min}`, :math:`t_{max}`], being:
+
+• :math:`t_{min}`: the time stamp of the first charge deposition of the event,
+• :math:`t_{max}`: defined considering that ``max_time`` =  :math:`t_{max}` - :math:`t_{min}`.
+
+.. image:: images/buffy/histogram_creation.png
+  :width: 800
+
+In this step, it also effectively pads with zeros in between separate signals. When these histograms (one for PMTs and another for SiPMs) are defined, they are sampled according to the binning of each sensor (``pmt_width`` and ``sipm_width``).
```
gonzaponte: How about "fills empty bins with zeros" or "restores empty bins by padding zeros in between separate signals"?

Also, we have to be careful with saying "one histogram for pmts and another one for sipms". My feeling is that it is not clear that we perform the syncronization using summed waveforms, but then we generate individual sensor waveforms from that.

Here I would clarify that these are waveforms produced by summing sensors. The generation of individual waveforms should be specified in the section "Synchronisation and trigger separation"

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+Sampling widths are included in the simulation parameters (``/MC/info``), and depends on the type of sensor and detector. Normally corresponds to 25 :math:`ns` for PMTs and and 1 :math:`\mu s`\ s for SiPMs.
+
+.. _Signal-Search:
+
+Signal Search
+::::::::::::::::::
+
+Once charge is distributed in previously defined histograms, the code searches for signal-like events.
```
gonzaponte:
```suggestion
Once the charge is distributed in the previously defined histograms, the code searches for signal-like events.
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+It takes PMT sum histogram and looks for the first value of the binned charge above a certain threshold (``trigger_threshold``), and defines the trigger time, :math:`t_{trigger}`.
```
gonzaponte:
```suggestion
It takes the PMT sum histogram and looks for the first value of the binned charge above a certain threshold (``trigger_threshold``), and defines the trigger time, :math:`t_{trigger}`.
```

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+
+.. note::
+  :math:`t_{min}` does not need to be at 0, since it is defined based on the first charge deposition, independently if it is above the ``trigger_threshold`` or not.
+
+.. _Trigg-Separation:
+
+Synchronisation and trigger separation
+:::::::::::::::::::::::::::::::::::::::
+
+Since the sensor width is different for PMTs and SiPMs, it is necessary to align and synchronise the clocks between their waveforms. Waveforms are then sliced then according to binning (``pmt_width`` and ``sipm_width``), trigger time and configured pre-trigger (``pre_trigger``).
```
gonzaponte: Here is where I think we need to specify that we generate individual PMT and SiPM waveforms based on the parameters obtained with the summed waveforms.

docs/source/buffy.rst
```diff
@@ -1,5 +1,143 @@
...
+If more than one trigger is found separated from each other by more than a buffer width, the nexus event can be split into multiple data-like triggers.
```
gonzaponte: buffer length instead of width to be consistent in the terminology?

gonzaponte: Also data-like waveforms instead of triggers?

gonzaponte (commented):
Much better. Now it's mostly grammar stuff, but I still think we need to clarify the process of obtaining individual waveforms from the summed waveforms.

docs/source/buffy.rst
halmamol: I already tried to find a way to add captions in read-the-docs and it's not that easy. I'll just add some text below or above.

docs/source/buffy.rst
halmamol: yes, they're! i've included some sentence along the text, not here!

docs/source/buffy.rst
halmamol: I've included some changes along the text to make this more clear!

halmamol:
Changes included! Before we fix the documentation, we need to sort out the issue with `max_time` since it is defined as an `int` (instead a `float`). I will work on it on IC.

gonzaponte:
I'm happy with this. Let's find someone else to have a second review :)

halmamol:
> I'm happy with this. Let's find someone else to have a second review :)

@gondiaz could give a hand on this, no need to be an expert on the city for the second review!

docs/source/buffy.rst
```diff
@@ -1,5 +1,145 @@
...
+   :widths: 50 40 120
+   :header-rows: 1
+
+   * - **Parameter**
+     - **Type**
+     - **Description**
+
+   * - ``max_time``
+     - ``int``
+     - Maximum duration of the event that will be taken into account starting from the first detected signal. All signals after that are lost. Must be greater than ``buffer_length``. If not, raises a warning and set ``max_time`` == ``buffer_length``
```
gondiaz: set --> sets?

docs/source/buffy.rst
gondiaz: I dont see any reason for it to be an `int` other than the argument (wrong) typing in the first of the links

docs/source/buffy.rst
```diff
@@ -1,5 +1,145 @@
...
+checks the time stamp of an event according to the sensors' response and defines histograms of charge distribution between [:math:`t_{min}`, :math:`t_{max}`], being:
+
+• :math:`t_{min}`: the time stamp of the first charge deposition of the event,
+• :math:`t_{max}`: defined considering that ``max_time`` =  :math:`t_{max}` - :math:`t_{min}`.
+
+.. image:: images/buffy/histogram_creation.png
+  :width: 800
+
+These histograms (one for PMTs and another for SiPMs) are defined by summing all individual sensors. This step restores also empty bins by padding zeros in between separate signals, and sample
+the histograms according to the binning of each type of sensor (``pmt_width`` and ``sipm_width``). Sampling widths are included in the simulation parameters (``/MC/info``), and depend on the type of sensor and detector.
```
gondiaz: `MC/configuration`

docs/source/buffy.rst
```diff
@@ -1,5 +1,145 @@
...
+
+Signal Search
+::::::::::::::::::
+
+Once the charge is distributed in the previously defined histograms, the code searches for signal-like events.
+It takes the PMT sum histogram and looks for the first value of the binned charge above a certain threshold (``trigger_threshold``), and defines the trigger time, :math:`t_{trigger}`.
+Waveforms are therefore defined for PMTs:
+
+• shifting the times of the charge histogram such that the first value over threshold (:math:`t_{trigger}`) falls at the time defined as ``pre_trigger``;
+• considering a specific length based on ``buffer_length``/``pmt_width``.
```
gondiaz: I think this sentence should be more concrete:
- set the length (in number of samples) as requested in the config parameters (`buffer_length/pmt_width`)

docs/source/buffy.rst
```diff
@@ -1,5 +1,145 @@
...
+
+.. note::
+  :math:`t_{min}` does not need to be at 0, since it is defined based on the first charge deposition, independently if it is above the ``trigger_threshold`` or not.
+
+.. _Trigg-Separation:
+
+Synchronisation and trigger separation
+:::::::::::::::::::::::::::::::::::::::
+
+Since the buffer length is different for PMTs and SiPMs, it is necessary to align and synchronise the clocks between their waveforms. Waveforms are then sliced then according to binning (``pmt_width`` and ``sipm_width``), trigger time and configured pre-trigger (``pre_trigger``).
```
gondiaz: I would say:
```
to align and synchronise the signals between waveforms
```

Also, remove second "then" in
```
Waveforms are then sliced then according to binning
```

gondiaz (changes requested):

docs/source/buffy.rst
halmamol: Done

gondiaz (approved):

gonzaponte (approved):
The inconsistency in the units of a variable has been addressed. Buffy is documented! Great job, @halmamol!

PR #12, opened 2022-10-24
Add Trude documentation

MiryamMV:
This PR adds the documentation for the city Trude.

All this info will eventually be here: https://pr012-sw-docs.readthedocs.io/en/latest/trude.html


docs/source/trude.rst
```diff
@@ -1,5 +1,109 @@
...
+     - ``string``
+     - Two options of subtracting the baseline: ``subtract_mode`` to subtract the mode and ``subtract_median`` to subtract the median.
+
+   * - ``number_integrals``
+     - ``int``
+     - Number of integrals to be performed in the waveform's time window. Typically this value is selected depending on the number of LED pulses in the waveform.
+
+   * - ``integral_start``
+     - ``int``
+     - Bin where the integrals start.
```
carmenromo: Maybe I would specify here that it is the bin where the first integral starts, so that then the other ones are integrated using the `integrals_period` by assuming the same separation between them.

docs/source/trude.rst
```diff
@@ -1,5 +1,109 @@
...
+
+ .. image:: images/trude/wf_intervals.png
+   :width: 850
+
+.. _Baseline subtraction of SiPM waveforms:
+
+Baseline subtraction of SiPM waveforms
+::::::::::::::::::::::::::::::::::::::
+
+Same procedure as described in :ref:`Baseline subtraction of SiPM waveforms` section of the *Irene* documentation with the option of using the mean instead of the mode of the waveform for the baseline subtraction.
```
carmenromo: This link does not go to `Irene`'s section. It is connected with the title of this part.

docs/source/trude.rst
```diff
@@ -1,5 +1,109 @@
...
+
+Spectrum histogram
```
carmenromo, line 106: This section does not have a link associated, like the other two. Is this on purpose?

docs/source/trude.rst
```diff
@@ -1,5 +1,109 @@
...
+   :widths: 50 40 120
+   :header-rows: 1
+
+   * - **Parameter**
+     - **Type**
+     - **Description**
+
+   * - ``min|max_bin``
+     - ``float``
+     - Lower/upper limit of the number of ACDs of the waveform to be considered for the spectrum.
```
carmenromo: Should `ADCs` be considered as unit? I'm wondering because of the guide instructions: https://github.com/next-exp/sw-docs/wiki#reviewer-guidelines

docs/source/trude.rst
carmenromo: And also Irene must be referenced using `:doc:name_city`.

docs/source/trude.rst
MiryamMV: Done!

docs/source/trude.rst
MiryamMV: I think I managed to fix this, let me know.

docs/source/trude.rst
MiryamMV: Fixed!

docs/source/trude.rst
MiryamMV: I would consider it as a unit. Already changed the font.

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
 Trude
 ==========
 
-.. caution::
-  This page is currently not created, we are working on it. If you would like to contribute on this documentation, reach `me <helena.almamol@gmail.com>`_!
+*From Germanic, Gertrud, hypocorism: female friend.*
+
+This city produces the light and dark spectrum of SiPMs for dedicated calibration runs. This is achieved by selecting regions in the SiPM waveforms where LED pulses are expected and regions which end 2 microseconds before, respectively and integrating the content within each region. The regions before LED pulses should only contain electronics noise and dark counts giving the zero external light approximation whereas those in time with the pulses will contain one or more detected photoelectrons. The waveform integrals are split into two groups: those with expected photoelectrons (light) and those without expected photoelectrons (dark). Each group produces a different spectrum.
```
joshgrocott: Is there any reason why 2 microseconds was the chosen time before the expected region? I understand this guarantees that the two types of region don't overlap, but why was 2 microseconds chosen instead of 3 microseconds for example? If this is unimportant then you can ignore this comment

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
...
+
+.. _Trude input:
+
+Input
+-----
+
+ * ``/Run/events``
+ * ``/Run/runInfo``
```
joshgrocott: Specific comments on the inputs might be nice for completeness, especially what is included in "runInfo"

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
...
+   :widths: 50 40 120
+   :header-rows: 1
+
+   * - **Parameter**
+     - **Type**
+     - **Description**
+
+   * - ``min|max_bin``
+     - ``float``
+     - Lower/upper limit of the number of :math:`ADCs` of the waveform to be considered for the spectrum.
```
joshgrocott, line 44: Could specify whether this applies to light spectrum or dark spectrum or both

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
...
+
+   * - ``bin_width``
+     - ``int``
+     - Width of the bins for the spectrum.
```
joshgrocott: Same as previous comment

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
...
+
+   * - ``proc_mode``
+     - ``string``
+     - Two options of subtracting the baseline: ``subtract_mode`` to subtract the mode and ``subtract_median`` to subtract the median.
+
+   * - ``number_integrals``
+     - ``int``
+     - Number of integrals to be performed in the waveform's time window. Typically this value is selected depending on the number of LED pulses in the waveform.
```
joshgrocott: Is this saying that two LED pulses, for example, would require two integrals? If so, you could give this as a specific example for clarity

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
...
+ * :ref:`SiPM waveforms baseline subtraction <SiPM waveforms baseline subtraction>`
+ * :ref:`Spectrum histogram <Spectrum histogram>`
+
+
+.. _Selection of integration limits:
+
+Selection of integration limits
+::::::::::::::::::::::::::::::::
+
+Given the config parameters, the first step in this city would be to selected the dark and light bins to compute the dark and light spectra.
```
joshgrocott: Just a spelling mistake/typo. "selected" should be changed to "select"

joshgrocott (commented):
Overall, I could understand everything well, most of my comments are just suggestions that may improve clarity :)

docs/source/trude.rst
MiryamMV: My guess is it is just to avoid the overlapping of both regions. @carmenromo is more experienced in using this code, perhaps she knows a better reason for this.

docs/source/trude.rst
MiryamMV: Maybe I didn't understand this comment, but Trude is used to compute the spectrum (light and dark) of a given pulsed waveform. Additionally, it can be used to compute just the dark spectrum if the input is a dark waveform, but it's not the main purpose. I don't find it necessary to include this clarification here.

docs/source/trude.rst
MiryamMV: Fixed!

docs/source/trude.rst
MiryamMV: Exactly! I just added an example.

docs/source/trude.rst
MiryamMV: Done.

docs/source/trude.rst
joshgrocott: No worries Miryam, I think it was me who read this part wrong!

docs/source/trude.rst
carmenromo: I think due to the pulse length 2 microseconds were enough to get the dark spectrum. But Andrew took the decision before we got into calibration :).

docs/source/trude.rst
```diff
@@ -10,9 +10,9 @@ This city produces the light and dark spectrum of SiPMs for dedicated calibratio
 Input
 -----
 
- * ``/Run/events``
- * ``/Run/runInfo``
- * ``/RD/sipmrwf``
+ * ``/Run/events``: list of the ``evt_number`` and the correspondent ``timestamp`.
+ * ``/Run/runInfo``: stores de ``run_number``.
```
carmenromo: I think here `de` is a typo.

MiryamMV: Fixed!

joshgrocott (approved):
Clear and concise. All my comments have also been addressed. Well done Miryam!

docs/source/trude.rst
halmamol: Maybe it could be included on the text that this value is hardcoded and chosen to avoid overlap of both spectrums. I think it's not explained and it's a fair question.

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
...
+
+   * - ``integral_start``
+     - ``int``
+     - Bin where the first integral of the set of ``number_integrals`` starts. The other integrals are performed using ``integrals_period`` assuming the same distance between the peaks.
+
+   * - ``integral_width``
+     - ``int``
+     - Number of bins to be considered for the integral starting at ``integral_start``. With ``integral_start`` these values are ideally chosen to select the whole pulse of the LED.
+
+   * - ``integrals_period``
```
halmamol, line 66: Does this have units? If so, I would add them.

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
...
+
+
+.. _Selection of integration limits:
+
+Selection of integration limits
+::::::::::::::::::::::::::::::::
+
+Given the config parameters, the first step in this city would be to select the dark and light bins to compute the dark and light spectra.
+
+The light bins will be centered in the light pulse(s) and the dark bins are chosen as an interval of ``integral_width`` bins which ends 2 microseconds before the light pulse begins, as shown here:
```
halmamol: Can you write microseconds in math mode (:math:`\mu s`)?

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
...
+
+ .. image:: images/trude/wf_intervals.png
```
halmamol, line 92: I think it will help when reading the documentation to have a better explanation of the parameters defined before and to which range, etc on the spectra correspond. You can check [Buffy](https://next-exp-sw.readthedocs.io/en/latest/buffy.html) documentation as a reference.

docs/source/trude.rst
```diff
@@ -1,5 +1,111 @@
...
+::::::::::::::::::::::::::::::::::::
+
+Same procedure as described in :ref:`Baseline subtraction of SiPM waveforms` section of the :doc:`irene` documentation with the option of using the mean instead of the mode of the waveform for the baseline subtraction.
+
+.. _Spectrum histogram:
+
+Spectrum histogram
+:::::::::::::::::::
+
+The last step would be the integration of the dark and light bins in order to obtain the respective spectrum histograms.
```
halmamol, line 109: Same comment as the one before. It would be good to mention which configuration parameters are used when creating the spectrum histogram.

halmamol (commented):
I've checked the PR, since @carmenromo is already out. Just few comments that could help on the documentation, and some beautification suggestions.

docs/source/trude.rst
MiryamMV: It is time units, the exact unit depends on the pulse.

docs/source/trude.rst
MiryamMV: I think I don't understand this comment. We use all of the config parameters listed above.

docs/source/trude.rst
MiryamMV: Done.

MiryamMV:
I added a couple more things to the documentation (including plots and changes in already existing plots). @joshgrocott feel free to take a general look to it and see if it makes sense. Any suggestions will be welcomed :)

docs/source/trude.rst
```diff
@@ -1,5 +1,167 @@
...
+     - ``string``
+     - Two options of subtracting the baseline. ``subtract_mode`` to subtract the mode and ``subtract_median`` to subtract the median.
+
+   * - ``number_integrals``
+     - ``int``
+     - Number of integrals to be performed in the waveforms time window. Typically this value is selected depending on the number of LED pulses to have in the waveform in the same time window.
+
+   * - ``integral_start``
+     - ``int``
+     - Bin where the integrals start.
```
joshgrocott: We found that in Phyllis that integral_start and integral_width need to be given in microseconds, but the waveforms used were in  (25 I think?) nanosecond time bins. I guess this would also be the case for trude? So would be worth mentioning in the documentation

joshgrocott (commented):
See comment regarding time bins for integral limits

docs/source/trude.rst
MiryamMV: Right, ``integral_start`` should be given in microseconds, but it's not the case for ``integral_width``.

PR #13, opened 2022-10-25
Document mau usage in S1 peak search

gonzaponte:
This PR documents the usage of the mau in the S1 peak search.
https://pr013-sw-docs.readthedocs.io/en/latest/irene.html


docs/source/irene.rst
```diff
@@ -161,7 +161,8 @@ Peak finding and matching of PMT and SiPM signals
...
 
-In order to optimize the peak search, PMT **CCWF**\ s are used as they have a higher sampling rate and therefore better time resolution. On top of that, these waveforms are PMT-summed to increase the signal-over-noise ratio [#]_. S1 and S2 signals are searched independently.
+In order to optimize the peak search, PMT **CCWF**\ s are used as they have a higher sampling rate and therefore better time resolution. The searches for S1 and S2 signals use slightly different waveforms. The search for S1 peaks uses the MAU-subtracted **CCWF**\ s. The PMT waveforms contain some low-frequency noise that can shift the PMT baseline locally by a small amount. The MAU (Moving Average Uindow) computes these small local variations of the baselines, which allows for a better peak finding algorithm. The parameter ``n_mau`` controls the size of this window. The S2 peak search uses the **CCWF**\ s directly.
```
halmamol: I wonder if we should define what is MAU before mention it, and not the other way around. Something like this: 

`"The search for S1 peaks uses the MAU *(Moving Average Uindow)* subtracted **CCWF**\ s. The PMT waveforms contain some low-frequency noise that can shift the PMT baseline locally by a small amount. The MAU computes these small local variations of the baselines, which allows for a better peak finding algorithm."`

halmamol (commented):
I have a couple of comments I can't add on the text:

- Maybe we should mention _MAU_ (in capital letters) in the definition of the variables, instead of _mau_. 
- Can you update the units in the config table to math format?

docs/source/irene.rst
gonzaponte: I have rewritten this part because it is not accurate to describe this whole thing as a MAU subtraction. In the process, MAU is defined the first time the term is used. Let me know what you think.

docs/source/irene.rst
```diff
@@ -161,7 +161,8 @@ Peak finding and matching of PMT and SiPM signals
...
 
-In order to optimize the peak search, PMT **CCWF**\ s are used as they have a higher sampling rate and therefore better time resolution. On top of that, these waveforms are PMT-summed to increase the signal-over-noise ratio [#]_. S1 and S2 signals are searched independently.
+In order to optimize the peak search, PMT **CCWF**\ s are used as they have a higher sampling rate and therefore better time resolution. The searches for S1 and S2 signals use slightly different waveforms. The PMT waveforms contain some low-frequency noise that can shift the PMT baseline locally by a small amount. The MAU (Moving Average Uindow) accounts for these small local variations of the baseline, allowing for a higher peak finding efficiency. The parameter ``n_mau`` controls the size of this window. The S1 peak search is performed by looking for samples of the PMT waveform that deviate from the MAU-generated baseline more than ``thr_mau``. On the other hand, the S2 peak search uses the **CCWF**\ s directly, without a MAU.
```
paolafer: What is a `Uindow`? Is it a spelling typo or is it on purpose?

gonzaponte: so, it should be MAW, not MAU, but in order to explain the acronym I put Uindow, but I don't have a strong opinion on this.

paolafer: So, is the acronym wrong? Or could we simply call it MAW?

gonzaponte: The acronym is MAW, but it's spelled MAU in the code.

paolafer: I see. It's a bit confusing, how about changing the code to the correct name `maw`?

gonzaponte: we can do that. Maybe there was a reason to call it mau, but I'm not aware of it.

halmamol (approved):
This PR documents the usage of MAW in the S1 peak search. It also triggered the rename of the variable in IC ([PR 828](https://github.com/next-exp/IC/pull/828)). Good job @gonzaponte !

paolafer (approved):
This PR clarifies the documentation, approved!

PR #16, opened 2023-01-12
Update production page

halmamol:
This PR updates the Official Production page with the tag versions and details for NEXT-100, and missing information about NEW. 

Updated version would like this: https://pr016-sw-docs.readthedocs.io/en/latest/production.html

### DIFF
--- a/docs/source/production.rst
+++ b/docs/source/production.rst
@@ -48,7 +48,7 @@ For the different productions the corresponding latest **tags** are:
      - ``NEXT_v1_05_02_NEXUS_v5_07_01_bkg_v9``
    * - **Calibration**
      - ``NEXT_v1_05_02_NEXUS_v5_07_01_bkg_v9``
-   * - **Background**
+   * - **NEW Background Model**
      - ``NEXT_v1_05_02_NEXUS_v5_07_01_bkg_v9``
 
 
@@ -75,20 +75,31 @@ Where the following **config** files have been used,
 
 NEXT-100
 ------------
-This sample is currently being produced. It is generated using
+A first production using fast simulations (:doc:`detsim`) has been produced using
 
 .. list-table::
    :widths: 40 60
    :header-rows: 0
 
-   * - **nexus**
-     - ``to be added``
-   * - **detsim**
-     - ``to be added``
+   * - **nexus: radiogenics + signal**
+     - ``v7_02_00``
+   * - **nexus: muons**
+     - ``v7_03_00``
+   * - **nexus: LTs and PSFs for Detsim**
+     - ``v7_01_00(+ G4 op bug)``
    * - **IC**
-     - ``to be added``
-   * - **Background Model**
-     - ``to be added``
+     - ``8e1de90c1cec5a1c3d25b35eca456f87f4f7e64c``
+   * - **IC: muons**
+     - `PR#823 <https://github.com/next-exp/IC/pull/823>`_
+   * - **NEXT-100 Background Model**
+     - ``v4``
+
+
+This production can be found in ``neutrinos1.ific.uv.es`` under the following path:
+
+.. code-block:: text
+
+ /data4/NEXT/NEXT100_prod/
 
 **Nexus macros** for the current production can be found on GitHub `here <https://github.com/gondiaz/NEXT100-0nubb-analysis/tree/main/nexus_job_templates/ft3>`_.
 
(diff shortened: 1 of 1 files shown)


gondiaz (approved):

docs/source/production.rst
```diff
@@ -75,27 +75,38 @@ Where the following **config** files have been used,
...
+     - ``v7_02_00``
+   * - **nexus: muons**
+     - ``v7_03_00``
+   * - **nexus: LTs and PSFs for Detsim**
+     - ``v7_01_00(+ G4 op bug)``
    * - **IC**
-     - ``to be added``
-   * - **Background Model**
-     - ``to be added``
+     - ``8e1de90c1cec5a1c3d25b35eca456f87f4f7e64c``
```
gonzaponte, line 91: We added a tag to IC I thought corresponded to the N100 preprod. It's [v.1.3.0](https://github.com/next-exp/IC/releases/tag/v1.3.0). Looking at the difference with this commit, it's almost irrelevant, except for one PR that changes the type of the event id column in the nexus tables so that it doesn't overflow. I don't remember if we decided to tag v.1.3.0 at [666647c](https://github.com/next-exp/IC/commit/666647c299b559163d47d9b4a17617edfea5886f) for other reasons or if this is a mistake. Maybe @gondiaz or someone else remembers?

docs/source/production.rst
```diff
@@ -75,27 +75,38 @@ Where the following **config** files have been used,
...
-   * - **nexus**
-     - ``to be added``
-   * - **detsim**
-     - ``to be added``
+   * - **nexus: radiogenics + signal**
+     - ``v7_02_00``
+   * - **nexus: muons**
+     - ``v7_03_00``
+   * - **nexus: LTs and PSFs for Detsim**
+     - ``v7_01_00(+ G4 op bug)``
```
gonzaponte, line 89: A note on what the bug is about and/or a link to where it is discussed in detail?

docs/source/production.rst
```diff
@@ -75,27 +75,38 @@ Where the following **config** files have been used,
...
    * - **IC**
-     - ``to be added``
-   * - **Background Model**
-     - ``to be added``
+     - ``8e1de90c1cec5a1c3d25b35eca456f87f4f7e64c``
+   * - **IC: muons**
+     - `PR#823 <https://github.com/next-exp/IC/pull/823>`_
```
gonzaponte, line 93: I assume that the PR is frozen. Otherwise, it doesn't give us a fixed version. I would add the commit hash in any case.

docs/source/production.rst
```diff
@@ -75,27 +75,38 @@ Where the following **config** files have been used,
...
 **Nexus macros** for the current production can be found on GitHub `here <https://github.com/gondiaz/NEXT100-0nubb-analysis/tree/main/nexus_job_templates/ft3>`_.
 
 Detsim *light tables* (**LTs**) and *point spread functions* (**PSFs**) can be found in ``neutrinos1.ific.uv.es`` in
 
 .. code-block:: text
 
-  /data5/users/gdiaz/NEXT100/kr83m/LightTables
+  /data4/NEXT/NEXT100_prod/LightTables/
 
 **Config** files for the rest of the production chain can also be found on `here <https://github.com/gondiaz/NEXT100-0nubb-analysis/tree/main/ic_processing/templates>`_ on Github.
```
gonzaponte: This is not exactly related to this PR, but we need to put the configs file in a central repository and not depend on a particular user. Maybe CERES will do the job. We can discuss it in the next sw meeting.

docs/source/production.rst
halmamol: @gondiaz can you provide me this?

docs/source/production.rst
halmamol: I agree, but so far it's where they're. Just in case we need to crosscheck something. I'll keep it mind.

docs/source/production.rst
gondiaz: https://bugzilla-geant4.kek.jp/show_bug.cgi?id=2471

docs/source/production.rst
gondiaz: I think this commit was after the productions, since we realized about tthe issue after the production was done. I have checked the production files and the event-id type is Int32, thus the v.1.3.0 was not in use.

Be warned that the commit hash could also be `3e81c90a6c66ae1ff32110423ed8f56ff5c0795f` which updates the activity database to `v3`. Nevertheless the activities used for the analysis are `v4`. I didn't took them from the database but directly from docDB.

docs/source/production.rst
gonzaponte: I would add this link to the docs page so we don't have to remember where it is (maybe a footnote?). Aside from that,  it's ready to be approved and merged.

PR #20, opened 2023-06-28
Update sipm waveform figure to match text in irene

gonzaponte:
https://pr020.readthedocs.io/en/latest/irene.html


halmamol (approved):
This corrects a typo on a figure on irene, and now matches what is written on the text.

bpalmeiro (approved):
Nice and simple correction, good catch

PR #23, opened 2024-01-11
Clarify pmaps hdf5 structure

gonzaponte:
A discussion in slack has brought up the point that the column names in the hdf5 pmap structure are not great. That made me realize that the documentation doesn't describe them properly either. This attempts to improve on that.

https://pr023-sw-docs.readthedocs.io/en/latest/

### DIFF
--- a/docs/source/hypathia.rst
+++ b/docs/source/hypathia.rst
@@ -37,7 +37,7 @@ Same as :doc:`irene`:
  * ``/Filters/empty_pmap``: flag for whether an event passed the empty pmap filter
  * ``/Filters/s12_indices``: flag for whether an event passed the s12 indices filter
 
- .. [1] The sensor index is the index in the corresponding database table (DataPMT or DataSiPM). In the case of PMTs, it coincides with the sensor ID.
+ .. [1] The sensor index is the index in the corresponding database table (DataPMT or DataSiPM in load_db.py). In the case of PMTs, it coincides with the sensor ID.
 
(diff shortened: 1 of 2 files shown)


kvjmistry:
For some reason when I click the link to [hypathia](https://pr020.readthedocs.io/en/latest/hypathia.html) it does not show your changes on the website. I am going off the git diff: 
- I would add the actual variable name in brackets (`nsipm`) just after the word "sipm id" etc. 
- If possible move the footnote so it is displayed above the next config heading as its hard to see when you have to scroll all the way to the bottom of the page. 
Otherwise, the description looks good to me!

halmamol:
> For some reason when I click the link to [hypathia](https://pr020.readthedocs.io/en/latest/hypathia.html) it does not show your changes on the website. 

The link you should be checking @kvjmistry is this one: https://pr023-sw-docs.readthedocs.io/en/latest/hypathia.html (on the PR description). Maybe that's the reason.

gonzaponte:
The link was wrong I updated it this morning, but forgot to comment it.

kvjmistry:
docs look good from my side!

kvjmistry (approved):

coreyjadams (approved):
Thanks for updating the docs. Oone thing I noticed is that in the actual DB, the PMTs and SiPMs are in the same table.  A good way to clarify this might be to include or link to the IC code that pulls the specified section of the table from the DB?

PR #24, opened 2024-01-18
Clarify nexus tag issue.

paolafer:
This PR clarifies the difference between the two nexus tags.
The name of the tags is also specified, because there's a typo  in the name of the folders (`10` instead of `01`).


paolafer:
We should also rename the `NEXT_v1_05_02_NEXUS_v5_07_10_bkg_v9` folder to `NEXT_v1_05_02_NEXUS_v5_07_01_bkg_v9` (both in the documentation and in the neutrinos server) and create a symbolic link with the old name. I have no permissions to do that, so maybe @jmbenlloch should do it.

paolafer:
The folders have been renamed, so this PR can be reviewed and merged!

gonzaponte (approved):
Crystal clear!

PR #28, opened 2024-02-15
Clarify that ROOT is not needed anymore

paolafer:
This PR clarifies the text of the NEXUS page regarding the ROOT version needed (up to a certain tag).
It also modifies a couple of other pages to solve compilation warnings and adds a .gitignore file to the repository, which was missing.

### DIFF
--- a/docs/source/NEXUS.rst
+++ b/docs/source/NEXUS.rst
@@ -33,9 +33,7 @@ Important information about NEXUS code:
 Tags
 ------------
 A list of the different NEXUS `tags <https://github.com/next-exp/nexus/wiki/Tags>`_ can be found on the Wiki.
-Starting from nexus tag v7_00_00, Geant4 version >= 11.0 is needed:
-
-For previous nexus tags,  ROOT >= 6.26 must be used.
+Starting from nexus tag v7_00_00, Geant4 version >= 11.0 is needed; for previous nexus tags,  ROOT >= 6.26 must be used.
 
  .. note::
    Geant4 version 10.7.x cannot be used with any nexus tag or even commit. For previous tags, Geant4.10.6 or Geant4.10.5 were used.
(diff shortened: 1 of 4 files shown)


docs/source/NEXUS.rst
```diff
@@ -33,10 +33,9 @@ Important information about NEXUS code:
 Tags
 ------------
 A list of the different NEXUS `tags <https://github.com/next-exp/nexus/wiki/Tags>`_ can be found on the Wiki.
-Starting from nexus tag v7_00_00, the following versions are needed:
+Starting from nexus tag v7_00_00, Geant4 version >= 11.0 is needed:
```
gonzaponte: The colon at the end doesn't make sense now. And I think the two sentences can now be in the same paragraph.

docs/source/NEXUS.rst
```diff
@@ -33,10 +33,7 @@ Important information about NEXUS code:
 Tags
 ------------
 A list of the different NEXUS `tags <https://github.com/next-exp/nexus/wiki/Tags>`_ can be found on the Wiki.
-Starting from nexus tag v7_00_00, the following versions are needed:
-
- * Geant4 version >= 11.0.
- * ROOT >= 6.26.
+Starting from nexus tag v7_00_00, Geant4 version >= 11.0 is needed; for previous nexus tags,  ROOT >= 6.26 must be used.
```
gonzaponte, line 36: Which Geant version should be used before v7?

paolafer: Not sure. Versions 11 and 10 should work (except for 10.7.x, as specified below), but before that I don't remember.

paolafer: In the blue box we just say that 10.5 or 10.6 can be used.

gonzaponte (approved):
Clear!

PR #32, opened 2024-04-25
Add references to the IC course material

gonzaponte:
https://pr032-sw-docs.readthedocs.io/en/latest/IC.html#a-tour-through-the-invisible-cities


jwaiton (approved):
Adds very useful course to the docs. Approved!

PR #33, opened 2024-05-16
Add path to new LTs and PSFs

gonzaponte:


docs/source/production.rst
```diff
@@ -96,6 +96,7 @@ Detsim *light tables* (**LTs**) and *point spread functions* (**PSFs**) can be f
 
 .. code-block:: text
 
-  /data5/users/gdiaz/NEXT100/kr83m/LightTables
+  /data5/users/gdiaz/NEXT100/kr83m/LightTables (for productions before May 2024)
```
paolafer: This first path doesn't seem to exist. There are tables under `/data5/users/gdiaz/NEXT100/LightTables/`, though. Are they the correct ones?

gonzaponte: Looks like it was wrong indeed. Fixed now.

paolafer (approved):
I've checked that the paths exist and do contain what promised. Approved!

PR #34, opened 2024-10-02
Adding clarification for blob centre calculation

jwaiton:
This is a minor adjustment to the isaura documentation, that alters the section **searching blobs position** to avoid confusion.

Before this change, it was easy to misinterpret the text as meaning that the hits within the voxels surrounding each extreme voxel were used to determine the blob centre, which isn't the case. **Only** the hits within the extreme voxel is used to determine the blob position.


gonzaponte (approved):
Good!

PR #37, opened 2025-05-12
Add warning to eutropia

jwaiton:
This PR adds a warning to the eutropia documentation that alerts the user of excessive memory usage/slowdown when the `xrange | yrange`, parameters are set to be too large. This in conjunction with a large number of `zbins` will inevitably cause computational issues and should be avoided if possible.

Hopefully eutropia can be rewritten to avoid this particular issue, but in the meantime a warning is appropriate.

The links can be found [here](https://pr37-sw-docs.readthedocs.io/en/latest/).

### DIFF
--- a/docs/source/eutropia.rst
+++ b/docs/source/eutropia.rst
@@ -102,12 +102,13 @@ following arguments:
      - ``float``
      - Bin size for :math:`\Delta x` and :math:`\Delta y`.
 
-.. warning:: 
+.. warning::
   A large 'PSF size' (defined by the ``xrange | yrange`` and 
-  ``zbins``) is discouraged. 
+  ``bin_size_xy``) in conjunction with a large number of ``zbins``
+  can cause excessive memory usage.
   To avoid this issue, ensure that your ``xrange | yrange`` do not
   exceed the PSF tails (~100mm for NEXT100), or use a smaller number
-  of ``zbins`` as a PSF is produced for each bin.
+  of ``zbins`` as a unique PSF is produced for each bin.
 
 .. _Eutropia workflow:
 
--- a/docs/source/eutropia.rst
+++ b/docs/source/eutropia.rst
@@ -103,10 +103,11 @@ following arguments:
      - Bin size for :math:`\Delta x` and :math:`\Delta y`.
 
 .. warning:: 
-  The values provided for ``xrange | yrange`` should not exceed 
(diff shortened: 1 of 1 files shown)


docs/source/eutropia.rst
```diff
@@ -102,6 +102,12 @@ following arguments:
      - ``float``
      - Bin size for :math:`\Delta x` and :math:`\Delta y`.
 
+.. warning:: 
+  The values provided for ``xrange | yrange`` should not exceed 
+  the PSF tails (~100mm for NEXT-100). Having large :math:`\Delta x`
```
gonzaponte: This has to be somehow combined with the bin size, since there is no such problem if you increase `bin_size` accordingly.

jwaiton: I phrased the next sentence the way I did:
```
Having large :math:`\Delta x` 
  and :math:`\Delta y` values causes excessive memory usage and 
  slowdown, especially when implementing a large number of ``zbins``
```
for this reason, as having one without the other (high `zbins`, low `xrange | yrange` or vice versa) is fine.

I can be more explicit. Something along the lines of:

The values provided for ``xrange | yrange`` should not exceed 
the PSF tails while implementing a large number of `zbins`, as
this causes excessive memory usage and slowdown.

gonzaponte: That's fine, but the bin size is also as important, I would perhaps introduce the term "PSF size", which is determined by `xyrange` and `bin_size`, and explain that if the PSF is large, it can cause excessive memory usage, particularly when using a large number of `zbins`, since we produce one PSF for each bin.

jwaiton: Ive added the concept of PSF size, and reworded it similar to your comment.

gonzaponte: Sorry for being so picky about this. My problem with it is that the PSF size is determined by 3 parameters: xrange, yrange and bin_size. Memory consumption is determined by the PSF size and the number of zbins. This is the bit that I find slightly misleading.

jwaiton: Apologies for taking some time to get back to this. It's better to be picky than have vague documentation :smile_cat: 

I've adjusted it to match as you said, although by `bin_size` I'm assuming that's `bin_size_xy`, and have named it as such :)

gonzaponte (approved):
Perfect!

